PR #5, opened 2019-07-25
specs for Options and Session

titusfortner:
These are the responsibilities I think we've agreed to, with a couple extra things.
Let's verify this is what we want.


nadvolod:
Looks gtm although very generic ;)

titusfortner:
@nadvolod yeah, this definitely isn't comprehensive, but even just looking at these should give us an idea of design requirements. I want for us to add to these and implement the "same things" in each of the bindings.

joshmgrant (approved):
I think this looks good.

PR #10, opened 2019-07-30
Fixed the code so that it compiles

nadvolod:
I refactored some tests and ignored others to get the build to compile. Also, did a bit of clean up as I saw fit. 
I created an issue here to fix the Ignored tests after we get the build green: https://github.com/saucelabs/simple_sauce/issues/9


mmerrell (approved):
This seems like a good start. A lot of open questions, but the only way to start answering them is to get something going

PR #14, opened 2019-07-30
Cleaning up

nadvolod:
Cleaned up a bunch of code so that it's better
Started implementing some of the specs for Sauce


java/src/main/java/com/saucelabs/simplesauce/Platforms.java
```diff
@@ -1,7 +1,17 @@
-package com.saucelabs.simple_sauce;
+package com.saucelabs.simplesauce;
 
 public class Platforms
 {
+	public static final String WINDOWS_10 = "Windows 10";
```
mmerrell, line 5: I'm starting to envision how we could move this to a kind of config file that would be generated based on a call to the Platform Configurator. I opened issue-17 about this. It's not a requirement for beta, I just wanted to throw that out there.

java/src/main/java/com/saucelabs/simplesauce/RemoteDriverInterface.java
```diff
@@ -6,6 +6,6 @@
 import java.net.MalformedURLException;
 
 public interface RemoteDriverInterface {
```
mmerrell, line 8: We should rename this to something else--part of the point of Interfaces is that users ideally wouldn't know or care that they're using an interface... this is the Object of Record that they care about. I also don't like "RemoteDriver", so we're going to have to come up with something else. It's not an emergency.

java/src/test/java/com/saucelabs/simplesauce/unit/DataCenterTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class DataCenterTest
+{
+    @Test
+    public void usWestDataCenterUrl_isCorrect()
+    {
+        assertThat(DataCenter.USWest,
+                IsEqualIgnoringCase.equalToIgnoringCase("https://ondemand.saucelabs.com/wd/hub"));
```
mmerrell, line 15: We shouldn't write unit tests just to verify strings like this... it's brittle, and doesn't really test anything important. Don't delete it, however--when we resolve issue-17, this test will start to become interesting

mmerrell (approved):
Once again, a lot of open questions, but a good start. I think the structure you're building here will lend itself well to enhancement and matching the specs we're fleshing out right now.

java/src/main/java/com/saucelabs/simplesauce/RemoteDriverInterface.java
nadvolod: @mmerrell Thanks for pointing that out. I'm happy to make changes to make my code as clean as possible. I try to avoid code rot and avoid Leblanc's Law. 
What this interface does is it instantiates a new RemoteWebDriver if it's the concrete implementation. And returns a fake for testing.
The reason why I added "_Interface_" to it is that it's really confusing as to which object I was interacting with. i was unable to look at the code and understand whether I'm using an interface or a concrete class. In C#, we put an "_I_" in front of our interface so that it's easy to look at the code and understand what the class is inheriting from. I don't know if that's a bad practice in Java? Also, maybe this confusion was a result of poor naming...
Side note. Our end uses would not know about this interface regardless. But the developers of the code would.

Some name ideas: RemoteDriverManager, RemoteDriverCreator, or something else?

java/src/test/java/com/saucelabs/simplesauce/unit/DataCenterTest.java
nadvolod: @mmerrell that's interesting. I'm curious as to why. The reason I've always tested this is for example: Let's pretend that someone decides to change the DataCenter.USWest url to something else, like Titus had "blah.blah:443". With pretty much all string concatenation, shit can always go wrong and you don't know how a new string will work where it's used. So usually, I validate my requirement strings so that if someone decides to change it, I have tests screaming that you broke something.
Btw, not disagreeing with you 👍  Simply trying to understand the logic as I'm sure you have more unit testing experience than I do. Let me know your thoughts

nadvolod:
@mmerrell see my comments above when you have a chance ^^

PR #15, opened 2019-07-31
Add makefile

joshmgrant:
The first draft of a Makefile for running tests in each language case


mmerrell:
I honestly don't love these things, but I understand that they work and that people know what they're here for. I don't have a better alternative at this point, so let's move ahead

joshmgrant:
I sort of agree with you @mmerrell. Makefiles can be clunky or, erm, lead to less than optimal processes. However, we can remove this later if the value isn't there.

PR #16, opened 2019-08-01
Implement Specs in Python

joshmgrant:
This is in-progress work to get the python bindings to comply with the initial (Ruby) specs.

### DIFF
--- a/python/simplesauce/session.py
+++ b/python/simplesauce/session.py
@@ -1,6 +1,17 @@
+import os
 from selenium import webdriver
 from selenium.webdriver.remote.remote_connection import RemoteConnection
 from simplesauce.options import SauceOptions
+
+
+SAUCE_USERNAME = os.getenv('SAUCE_USERNAME', None)
+SAUCE_ACCCESS_KEY = os.getenv('SAUCE_ACCESS_KEY', None)
+
+us_ondemand = 'ondemand.saucelabs.com'
+eu_ondemand = 'ondemand.eu-central-1.saucelabs.com'
+
+US_SAUCE_DC_URL = 'https://{}:{}@{}/wd/hub'.format(SAUCE_USERNAME, SAUCE_ACCCESS_KEY, us_ondemand)
+EU_SAUCE_DC_URL = 'https://{}:{}@o{}/wd/hub'.format(SAUCE_USERNAME, SAUCE_ACCCESS_KEY, eu_ondemand)
 
 
 class SauceSession():
--- a/python/simplesauce/session.py
+++ b/python/simplesauce/session.py
@@ -1,6 +1,6 @@
 from selenium import webdriver
 from selenium.webdriver.remote.remote_connection import RemoteConnection
-from options import SauceOptions
+from simplesauce.options import SauceOptions
 
 
 class SauceSession():
--- a/python/simplesauce/session.py
+++ b/python/simplesauce/session.py
@@ -11,15 +11,18 @@ us_ondemand = 'ondemand.us-west-1.saucelabs.com'
 eu_ondemand = 'ondemand.eu-central-1.saucelabs.com'
 
 US_SAUCE_DC_URL = 'https://{}:{}@{}/wd/hub'.format(SAUCE_USERNAME, SAUCE_ACCCESS_KEY, us_ondemand)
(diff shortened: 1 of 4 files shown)


python/simplesauce/session.py
```diff
@@ -0,0 +1,85 @@
...
+    @data_center.setter
+    def data_center(self, data_center):
+        if data_center.lower() == 'eu':
+            self._data_center = data_center
+            self.remote_url = EU_SAUCE_DC_URL
+        elif data_center.lower() == 'us':
+            self._data_center = data_center
+            self.remote_url = US_SAUCE_DC_URL
+        else:
+            raise KeyError("Invalid Data Center value, please select from 'us' or 'eu'")
```
mmerrell, line 66: Should we include either the option to add a custom url, or at least add the Headless endpoints? we might need to have some notion of "VM-US", "VM-EU", "Headless" (however we'd abbreviate that), etc... this might merit a discussion.

python/simplesauce/session.py
```diff
@@ -0,0 +1,85 @@
...
+
+US_SAUCE_DC_URL = 'https://{}:{}@{}/wd/hub'.format(SAUCE_USERNAME, SAUCE_ACCCESS_KEY, us_ondemand)
+EU_SAUCE_DC_URL = 'https://{}:{}@{}/wd/hub'.format(SAUCE_USERNAME, SAUCE_ACCCESS_KEY, eu_ondemand)
+
+
+class SauceSession():
+
+    def __init__(self, data_center='us', username=None, access_key=None, options=None):
+
+        # TODO: flesh this out
```
mmerrell, line 21: We should flesh this out in terms of a) figuring out what the options are, and b) making sure we're unified across the languages in how we implement them.

python/simplesauce/session.py
```diff
@@ -0,0 +1,85 @@
...
+        self._username = username if username else SAUCE_USERNAME
+        self._access_key = access_key if access_key else SAUCE_ACCCESS_KEY
+        self._data_center = data_center
+
+        if data_center.lower() == 'eu':
+            self.remote_url = EU_SAUCE_DC_URL
+        elif data_center.lower() == 'us':
+            self.remote_url = US_SAUCE_DC_URL
+        else:
+            raise KeyError("Invalid Data Center value, please select from 'us' or 'eu'")
```
mmerrell, line 33: See comment below--we need to make sure we add the Headless data center, and keep this somewhat open for new endpoints. Could the "else" clause validate the data_center parameter as a URL, try to open a session at the URL as an endpoint, and then raise the KeyError if it doesn't work? Is that non-Pythonic?

python/tests/test_options.py
```diff
@@ -0,0 +1,80 @@
...
+
+        assert sauce.browserName == 'chrome'
+        assert sauce.browserVersion == 'latest'
+        assert sauce.platformName == 'windows 10'
+
+    def test_accepts_browser_version_platform_name(self):
+        sauce = SauceOptions('Firefox', '67.0', 'windows 10')
+
+        assert sauce.browserName == 'Firefox'
+        assert sauce.browserVersion == '67.0'
```
mmerrell, line 17: We need to figure out across the languages how we're going to be able to support the valid browser versions outside of pushing changes to these repos every time it changes. I wonder if we could create a dead-simple call to the Platform Configurator to create a cache of valid browser versions, and create defaults for them. This is a larger discussion. Ideally it would live in a "common" area of the project, usable by all languages...

mmerrell (approved):
This is a good first cut--I think we have some decisions to make about how all of these things will be implemented, but those things don't need to be answered in order to merge this in.

python/simplesauce/session.py
joshmgrant: I can imagine this would be helpful. For now, I'd say we only set fixed defaults but allow customization later.

python/tests/test_options.py
joshmgrant: This is probably the right way to go. In this case, I'm naively validating that what version is passed in is kept true by the `SauceOptions` instance. Integrating the Platform Configurator somehow to get valid/invalid/up to date versions of browsers would be excellent, and could definitely live in a common area.

PR #21, opened 2019-08-07
Adding travis

joshgrantsauce:
First draft at adding CI via Travis. Two notes:

- If this PR is merged, we will have Travis support
- This is a first draft of how to execute multiple test suites concurrently. The log output is suboptimal due to how `make` parallelizes processes so this may be improved in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,13 @@
 sudo: true
 
-java: &java
-  language: java
-  jdk: oraclejdk8
-  before_script: cd java
-  env:
-    global:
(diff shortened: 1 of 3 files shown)


nadvolod:
@joshgrantsauce looks good tome man. Looks like it's missing sauce credentials for UI automation. but other than that, is it ready for merging?

joshmgrant:
@nikolay-advolodkin Yep, that's the missing piece here is managing credentials. Not sure how we want to do this (a dedicated user for testing? use our own? store via Travis? etc)

PR #23, opened 2019-08-09
Pypi package

joshmgrant:
Added necessary bits to push up to [PyPI](https://pypi.org/) and allow for easy installation via pip.


nadvolod:
@joshmgrant looks like the Quality Review is failing as well as CI. Are you able to see the Code Quality report to analyze it and see why the quality is failing? THe tool has default rules turned on which for the most part are excellent. But it's up to the language master to set what they think makes and doesn't make sense.

joshmgrant:
@nikolay-advolodkin thanks for the nudge on the Codacy checking! It looks like it does have some python formatting set up somehow, so I pushed a fix.

nadvolod:
@joshmgrant you should have access to see what those problems are, to fix them and to configure the rules if they don't make sense. Let me knwo if you need a quick demo of how to do this?

PR #33, opened 2019-09-10
1st project in .NET

nadvolod:


TheTestLynx (approved):
Is this TDD? (sauce con class has no content, but I feel it is intended this way for now)

PR #37, opened 2019-10-02
Add ability to pass SauceOptions to SauceSession

nadvolod:


nadvolod:
Please take a look at my naming conventions for Edge, I went with something like this:
```withEdge13_10586()```, but maybe I should just do ```withEdge13()```

File: java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java [hunk 1]
```diff
@@ -27,13 +25,13 @@ public SauceOptions withSafari()
     {
         return withMacOsMojave();
     }
-    public SauceOptions withSafari(String version)
+    public SauceOptions withSafari(final String version)
     {
-
-        if (version.isEmpty()) { version = "latest"; }
+        String _version = version;
```
nadvolod, line 30: @joshmgrant so I see how this is better than what I had before. However, what prompted the change to have another local **_version** variable? I liked this old code on line 33:
```if (version.isEmpty()) { version = "latest"; }```
But would it be better to do ```if (version.isEmpty()) { version = browserVersion; }```
Since `browserVersion` is set to "latest" already

joshmgrant: This would be fine either way.

PR #43, opened 2019-10-18
Fixed Edge versions to be withEdgeN versus withEdgeN_nnnn

nadvolod:
So rather than doing withEdge15_12345() we can now do withEdge15() and other Edge versions

### DIFF
--- a/java/src/main/java/com/saucelabs/simplesauce/IEVersion.java
+++ b/java/src/main/java/com/saucelabs/simplesauce/IEVersion.java
@@ -1,0 +1,6 @@
+package com.saucelabs.simplesauce;
+
+public enum IEVersion {
+    ;
+    public static final String _11 = "11.285";
+}
--- a/java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
+++ b/java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
@@ -3,6 +3,7 @@ package com.saucelabs.simplesauce;
 import lombok.Getter;
 import lombok.Setter;
 import org.openqa.selenium.chrome.ChromeOptions;
+import org.openqa.selenium.edge.EdgeOptions;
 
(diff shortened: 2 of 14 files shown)


java/src/main/java/com/saucelabs/simplesauce/IEVersion.java
```diff
@@ -0,0 +1,6 @@
+package com.saucelabs.simplesauce;
+
+public enum IEVersion {
+    ;
```
TheTestLynx, line 4: ;

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -0,0 +1,148 @@
...
+        operatingSystem = "Windows 8";
+        return this;
+    }
+
+    public SauceOptions withWindows7() {
+        operatingSystem = "Windows 7";
+        return this;
+    }
+
+    public SauceOptions withEdge() {
```
TheTestLynx, line 94: If this method is public maybe we should add the latest browserVersion too and change it if specified? Or have it private?

TheTestLynx (commented):
Please check also the other comments about empty string and vista

java/src/main/java/com/saucelabs/simplesauce/IEVersion.java
nadvolod: Yea, unless I missed something obvious, this is how the compiler forces me to do it in Java 👎 @joshmgrant does that look right to you?

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
nadvolod: @TheTestLynx can you elaborate a bit more? Are you saying to be able to pass in a `edgeVersion` into this method?

Issue #46, opened 2019-10-18
Fix some or all TODOs in Java bindings


nadvolod:
There are a bunch of minor TODO comments in the Java bindings that can be fixed to make the code cleaner and more understandable

picimako:
Hi @nadvolod,
I wanted to work on this issue, however it seems that there are no TODO comments in the Java bindings, at least I couldn't find any.  :)

nadvolod:
Hey @picimako super happy that you're looking to help out here 👍 Yea, we went through quiet a refactoring period that probably rendered this issue and other's obsolete. Let me clean up here and see where we can take your help

nadvolod:
@picimako did you want to take a stab at https://github.com/saucelabs/sauce_bindings/issues/144 or https://github.com/saucelabs/sauce_bindings/issues/145 ?

picimako:
@nadvolod I'm happy to help. :)
I'm looking into the issues you mentioned.

PR #47, opened 2019-10-21
Fix duplicate withEdge() methods

jaeheonshim:
I saw that in the Java bindings the methods regarding the Edge browser were duplicated unnecessarily for each version. To make the code cleaner, I created an enumeration set for each version, and removed all but one method that takes in the version as a parameter.

Issue #46

### DIFF
--- a/java/src/main/java/com/saucelabs/simplesauce/SauceSession.java
+++ b/java/src/main/java/com/saucelabs/simplesauce/SauceSession.java
@@ -17,13 +17,15 @@ import org.openqa.selenium.safari.SafariOptions;
 import java.net.MalformedURLException;
 
 public class SauceSession {
-    @Getter @Setter public final String sauceDataCenter = DataCenter.USWest;
+    @Getter
+    @Setter
+    public final String sauceDataCenter = DataCenter.USWest;
     private final EnvironmentManager environmentManager;
     public SauceTimeout timeouts = new SauceTimeout();
 
     //todo there is some weird bug when this is set to Linux, the session can't be started
-	private String operatingSystem = "Windows 10";
-	private String browserName = "Chrome";
+    private String operatingSystem = "Windows 10";
+    private String browserName = "Chrome";
     private final String sauceOptionsTag = "sauce:options";
     private ChromeOptions chromeOptions;
     private FirefoxOptions firefoxOptions;
@@ -35,7 +37,9 @@ public class SauceSession {
     private WebDriver webDriver;
     private EdgeOptions edgeOptions;
     private InternetExplorerOptions ieOptions;
-    @Getter @Setter public String sauceLabsUrl;
+    @Getter
+    @Setter
+    public String sauceLabsUrl;
 
     public SauceSession() {
         sauceSessionCapabilities = new MutableCapabilities();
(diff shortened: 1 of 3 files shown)


java/src/main/java/com/saucelabs/simplesauce/SauceSession.java
```diff
@@ -223,42 +223,12 @@ public SauceSession withMacOsXYosemite() {
         browserName = "Safari";
         return this;
     }
-    //TODO notice the duplication below with edge.
+    //FIXED notice the duplication below with edge. Used enumeration
     //Maybe could be moved to a separate class so we can do withEdge().16_16299();
     //Or withEdge().version(EdgeVersion.16_16299);
-    public SauceSession withEdge16_16299() {
+    public SauceSession withEdge(EdgeVersion version) {
```
nadvolod, line 218: That's an interesting solution. I'd like to get your feedback on this:
So, the idea here with these bindings is to have as much static typing as possible. Meaning, we wanted to do stuff like: `withEdge16()` and `withEdge17()` and so on. That way, the user isn't left guessing about what all of the supported Edge versions are.
Your solution solves the same problem a different way, using an `enum`. 
[Here](https://github.com/saucelabs/simple_sauce/pull/43/files#diff-737dcd8dd9072c63fbbd35c93414eabfR102-R111) I actually fixed the naming convention of the Edge versions to `withEdge15()` rather than 'withEdge15_5757'. I tend to like the new name better. In which case, using an Enum will not work.
What are your thoughts?

joshmgrant: Personally I was thinking Enums could be helpful somewhere here. We might be able to default to a version too with an overloaded method.

jaeheonshim: I've updated my pull request to update the naming convention as per nadvolod's suggestions. You can view the updates [here](https://github.com/saucelabs/simple_sauce/pull/47/commits/3168ce4295b7b4219081fd885afdbe7ef3900328). The request still uses enumerations, but changes the version name.

nadvolod: @joshmgrant hey bud, can you elaborate? Are you liking the direction that @jaeheonshim is going you're saying? Or something else?

java/src/test/java/com/saucelabs/simplesauce/unit/EdgeTests.java
```diff
@@ -1,51 +1,19 @@
...
-        String actualBrowserSetInConfig = fakeSauceSession.sauceSessionCapabilities.getVersion();
-        assertEquals("14.14393", actualBrowserSetInConfig);
-    }
-    @Test
-    public void withEdge13_10586_returnsBrowserVersion13_10586() {
-        fakeSauceSession.withEdge13_10586();
-        fakeSauceSession.start();
-        String actualBrowserSetInConfig = fakeSauceSession.sauceSessionCapabilities.getVersion();
-        assertEquals("13.10586", actualBrowserSetInConfig);
+    public void withEdgeReturnsCorrectBrowserVersion() {
```
nadvolod, line 10: @mmerrell how do you feel about this test? I think that we need a separate test for every Edge version because it's the requirement that we support each one. Or at least, maybe the supported edge versions can come from a collection rather than the actual enum?

nadvolod:
@jaeheonshim hi, are you interested in merging this in? If so, please respond back relatively soon.
Seems like you need to update this branch with master. Run tests locally and see what's broken. Fix tests. Then push those changes up.
Also, the EdgeTest should be updated to test actual strings through a data provider rather than testing the enum against itself. I believe the test will never fail as a result.

PR #50, opened 2019-10-24
basic idea is present

joshmgrant:
Thinking of adding additional features, here's a way to capture test statuses on Sauce.


joshmgrant:
No longer needed in light of work at hackathon.

PR #51, opened 2019-10-24
two example classes

joshmgrant:
Code-based examples are helpful for folks getting started, so here are some starting points in fully executable Java code.


nadvolod:
@joshmgrant looks good except the failing Codacy checks. Looks like some unused imports

nadvolod:
@joshmgrant same thing here with the unused imports

joshmgrant:
@nikolay-advolodkin why was the `getDriver()` getter removed in SauceSession? Seems like a good method to have.

nadvolod:
@joshmgrant because I think it had no usages. Creating the SauceSession automatically returns it so that method stopped being used. PS, I don't use nikolay-advolodkin  anymore, use nadvolod please :)

java/src/main/java/com/saucelabs/simplesauce/SauceSession.java
```diff
@@ -65,23 +64,25 @@ public SauceSession(SauceOptions options, RemoteDriverInterface remoteManager, E
         this.environmentManager = environmentManager;
     }
 
+    public WebDriver getDriver() {
```
nadvolod, line 67: @joshmgrant I think that if you use Lombok, we don't need this. 
Keep this `@Getter public WebDriver webDriver;` but make it `public` and then with Lombok, the user should be able to just do `sauce.getwebDriver()` and it will automatically work with magic.

nadvolod: this can be changed I think

nadvolod (changes requested):

joshmgrant:
Closing in favour of #75

PR #58, opened 2019-10-29
started main readme

joshmgrant:
Tried to cover all the bases. We will need to update the language-specific docs as well.


nadvolod (approved):
This is great man! 👏

PR #61, opened 2019-10-30
Py w3c update

joshmgrant:
Updates to correctly send W3C capabilities to Sauce Labs in the Python case


joshmgrant:
@nikolay-advolodkin this PR looks good, and the codacy issues that are left can be ignored/disabled. I just can't seem to do that?

nadvolod (approved):

PR #63, opened 2019-11-05
Java instructions

joshmgrant:
Adding detail to the README for how to get started with Simple Sauce installation and testing

### DIFF
--- a/java/README.md
+++ b/java/README.md
@@ -67,3 +67,55 @@ Edge version 15
 SauceOptions options = new SauceOptions().withEge15();
 webDriver = new SauceSession(options).start();
 ```
+
+## Installation
+
+Simple Sauce will be available via Maven soon but for now please follow the [Building](#building) instructions for local development.
+
+## Building
+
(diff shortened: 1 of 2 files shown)


java/README.md
```diff
@@ -67,3 +67,55 @@ Edge version 15
...
+```bash
+mvn install
+```
+
+in the `java/` directory. This will allow you to import Simple Sauce as a Maven dependency in other local projects on your laptop as a snapshot of the state based on the commit that you build from. To add this reference via Maven, add
+
+```xml
+<groupId>com.saucelabs</groupId>
+<artifactId>simple_sauce</artifactId>
+<version>1.0-SNAPSHOT</version>
```
nadvolod, line 108: For this part, does it make sense to mention that this is dynamic and how to retrieve that value?

nadvolod (changes requested):

java/README.md
joshmgrant: That sounds good, I'll add it in.

PR #69, opened 2019-11-14
Fix broken W3C wiki link

richard-sl:
Small fix it looks like the direct link changed for the wiki page on W3C


joshmgrant (approved):
Love it, thanks for fixing the broken link!

PR #70, opened 2019-11-15
Bug fix for stop()

nadvolod:
I've introduced a bug fix for the `stop()`.
I believe that I found the simplest solution that makes sense. By allowing the user to pass in their `webDriver` instance to the `stop()`


nadvolod:


```diff

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/SauceDriverImpl.java  1
         

Complexity decreasing per file
==============================
+ java/src/test/java/com/saucelabs/simplesauce/acceptance/SauceSessionAcceptanceTest.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/simple_sauce/simple_sauce/pullRequest?prid=4508295&bid=15193147)

nadvolod:


```diff

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/SauceDriverImpl.java  1
         

Complexity decreasing per file
==============================
+ java/src/test/java/com/saucelabs/simplesauce/acceptance/SauceSessionAcceptanceTest.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/simple_sauce/simple_sauce/pullRequest?prid=4508295&bid=15193147)

nadvolod:


```diff

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/SauceDriverImpl.java  1
         

Complexity decreasing per file
==============================
+ java/src/test/java/com/saucelabs/simplesauce/acceptance/SauceSessionAcceptanceTest.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/simple_sauce/simple_sauce/pullRequest?prid=4508295&bid=15193147)

nadvolod:


```diff

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/SauceDriverImpl.java  1
         

Complexity decreasing per file
==============================
+ java/src/test/java/com/saucelabs/simplesauce/acceptance/SauceSessionAcceptanceTest.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/simple_sauce/simple_sauce/pullRequest?prid=4508295&bid=15193147)

java/src/main/java/com/saucelabs/simplesauce/SauceSession.java
```diff
@@ -137,10 +137,10 @@ private void tryToCreateRemoteWebDriver(String sauceLabsUrl) {
             throw new InvalidArgumentException("Invalid URL");
         }
     }
-    public void stop()
+    public void stop(WebDriver driver)
```
joshmgrant, line 140: @nadvolod requiring a `null` argument (at least) in every call of `sauce.stop()` feels bad to me. Can we add an overloaded method that does not take an argument? Something like

```java
public void stop(WebDriver driver) { 
     // logic 
}
public void stop() {
    // logic that might reuse the above method
}
```

I'm hoping these two ideas encourage folks to use the "right" pattern for when they use a SauceSession object.

nadvolod: @joshmgrant why are you assuming that someone would need to pass in `null`? I might be missing something, it happens, but here are the 2 use cases that I see possible:
1. This is the most common use case by far. And it makes it mandatory to pass in the driver that you got from the start()
```
driver = new SauceSession().start();
driver.doStuff();
session.stop(driver);
```
2. This use case I can't imagine why someone would do this for test automation, but technically it's possible. It's basically starting a Sauce Session without doing anything with a driver. Literally just starting a session.
```
session = new SauceSession();
session.start();
session.stop(the problem is here, no driver to pass in because they didn't do anything with a driver);
```
With that said, I can see the argument for what you posted based on scenario 2. But then we open our API up to users doing `stop()` instead of `stop(driver)` by mistake and causing problems for themselves since it's almost always mandatory to do the latter. And this is the most common use case. Hence, I wonder if we are optimizing for the exception, rather than the rule?
Please let me know your thoughts :)

joshmgrant: @nadvolod There is a third use case, assuming that there's a getter for the driver member of the SauceSession object, that would look like this:

```java
SauceSession sauce = new SauceSession();
sauce.start();
// actions with the sauce object
sauce.stop();
```

This use case makes sense if you consider that the driver used within the SauceSession instance is only created to be passed into a page object-like object:

```java
SauceSession sauce = new SauceSession();
sauce.start();
PageObject page1 = new LoginPage(sauce.getDriver());
PageObject page2 = new AccountPage(sauce.getDriver());
// actions with page1 and page2
sauce.stop();
```

Here we don't even need to create an intermediate driver instance. We just pass around `sauce`. 

Of course you could follow the pattern of

```java
SauceSession sauce = new SauceSession();
WebDriver driver = sauce.start();
PageObject page1 = new LoginPage(driver);
PageObject page2 = new AccountPage(driver);
// actions with page1 and page2
sauce.stop(driver);
```

but this seems somewhat awkward particularly if `driver` and `sauce` are used in multiple files. However, looking at this I feel like both options are good and should be supported.

nadvolod: @joshmgrant thanks for sharing, I see that now. Let me know your thoughts on this:
At some point, regardless of this 
```
PageObject page1 = new LoginPage(sauce.getDriver());
```
the user would need to do `webDriver = sauce.getDriver()` in their page objects, but most likely the `BaseTest`
The scariest thing for me out of all of this is for the user to interact with the driver like this:
```
sauce.getDriver().findElement();
//and more interactions like this
```
This will not only make the usage of Easy Sauce annoying as it forces an extra method call on every action, but will also introduce vendor lock in as now `sauce` is our services that's used all over the code. As opposed to having a single place to just set the driver as such `WebDriver driver = sauce.start();` or even as such `WebDriver driver = sauce.getDriver();`

In which case, does that make the 3rd option moot?

joshmgrant: The easy fix here is to have a `WebDriver driver` member in their Page Object classes. Then in the constructor the driver gets cleanly assigned to the instance driver via `sauce.getDriver()`. Easy :)

I do think both patterns are valid at this point, and we should keep two methods for the `stop()` method.

nadvolod: @joshmgrant so you're saying this?
```
public class PageObject
{
    public PageObject(WebDriver driver){
    this.driver = driver;
    }
}
```
Usage
```
page = new PageObject(sauce.getDriver());
page.doStuff();
sauce.stop();
```

joshmgrant: Exactly. This is a good practice and we should encourage it.

nadvolod: @joshmgrant agreed about the practice. However, this will fail on the `sauce.stop()` because you have to pass in the driver for the stop to work correctly. That's the bug you pointed out in the first place :) Not sure if I'm missing something here, but there is no way that `stop()` will work if the user sets their own driver variable because now they've got a different instance. So, if at any point the user does this
```
this.driver = sauce.getDriver();
```
then they have to use `stop(WebDriver driver)`. There's no way around it.
The only time they won't have to use that method is if they do this for every single driver command:
```
sauce.getDriver().findElement();
sauce.getDriver().get();
...
```
But that's really verbose and binds the user to Sauce Labs implementation. Hence, I don't believe that there is any valid use case for `stop()`. It seems to be only a source of confusion? Does that make sense or am I missing something?

nadvolod:


```diff

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/SauceDriverImpl.java  1
         

Complexity decreasing per file
==============================
+ java/src/test/java/com/saucelabs/simplesauce/acceptance/SauceSessionAcceptanceTest.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/simple_sauce/simple_sauce/pullRequest?prid=4508295&bid=15193147)

PR #89, opened 2019-12-20
[py] update logic to match Ruby

titusfortner:
@joshmgrant based on our earlier conversation...

Mostly this is making use of the meta-programming to allow all of the values to be passed in during construction or with get/set.

But I added tests to Ruby & Python and created implementation to bring these to parity.

PR #92, opened 2019-12-23
Java updates

titusfortner:
@nadvolod first off, don't freak out; the removal of enums / strict method setters is temporary, we can pull that code back in as we need it. :)

This brings Java in parity with Ruby & PR #89 for Python. I think this PR is sufficient for beta for Java. Which isn't to say that we can't continue to put in the extra features before a release. I'm just trying to get us on equal footing first.

There was a fair bit in here that we were duplicating from Selenium and some overhead that I think is useful but not required in the short term. The big thing with respect to the Options class is that we are not doing what Selenium Options classes do. They maintain a Map of key value pairs and create custom getters/setters that read and write to that Map. This implementation is using lombok to set native Java field values, which means that it does not allow any non-known capabilities, but also, it means nothing has to be hard coded. The heavy lifting is managed with reflection in the `toCapabilities` method. Also, the `setCapability` method is currently private. I think it should stay private, but I'm open to differing opinions here, I'll discuss it more below.

What's Next (for ruby/python as well, but specifically looking at it from a Java standpoint):
1. Make it easy for users to know and choose acceptable values 
    * Enums for everything with a known and limited set of options (**discussion below**)
    * Static methods: I don't understand the purpose of these if we do the enums correctly.
        * Is this effectively an alternative to enum? (I'm struggling with enums, so considering this as an alternative might be useful)
        * Is this mostly to support a Fluent pattern? (I don't love it, but if we want it we just need to make sure it is organized)
        * Does the implementation of this PR with all of its getters and setters affect how we think about these? (e.g. mixing `withEdge` and `setBrowserVersion` would give different results depending on order called)

2. Be smarter about Mac/Safari and IE/Windows combination validations (@joshmgrant this is the safari conversation we had on the Python PR; the default right now is to require users to be more explicit first before we do convenience things that might result in more maintenance costs)

3. Mocking the tests (I figured out ENV mocks, but didn't get far enough along with mockito to figure out how to ensure that the driver is getting sent the right things). It's not close enough to what I'm used to in Ruby for it to be trivial, so I just made sure the tests actually passed on my Sauce account. Not ideal; but honestly the session stuff right now is the least concerning bit of this code to me.

ok, enums.

I've tried several different ways to get this working and can't figure out a way to do it that doesn't break the current specs... 

The main use case I'm interested in is being able to pull in serialized information (YAML, JSON, Property, XML) as a lookup rather than needing to create a large conditional for everything. In Ruby & Python we can initialize the options object with a Hash obtained from one of these sources. I don't think we should do that in Java. In Java I implemented a `setCapabilities` method that takes a `Map` instance. With the current code, what I need to do works. Using an enum makes it impossible to set a String value, which means we either have to maintain the Map that I said above I don't want to do, or we need to be able to translate things. We can hard code in these translations, but that is a huge maintenance cost. We can try to use reflection to do the translations, but after a couple different tries I couldn't get it to work.

For reference, I think the implementation we want would allow this:
```
getCapability("platformName")              // => null

setPlatformName(PLATFORM.MAC_SIERRA)
getPlatformName                            // => PLATFORM.MAC_SIERRA
getCapability("platformName")              // => "macOS 10.12"

setCapability("platformName", "Windows 10")
getCapability("platformName")              // => "Windows 10"
getPlatformName                            // => PLATFORM.WIN_10
```

Anyway, please provide feedback, I'm going to keep thinking the thoughts on the next step here.


nadvolod:


```diff

Issues
======
- Added 7
           

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java  6
- java/src/main/java/com/saucelabs/simplesauce/SauceDataCenter.java  1
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
- java/src/main/java/com/saucelabs/simplesauce/TestResult.java  1
         

Complexity decreasing per file
==============================
+ java/src/main/java/com/saucelabs/simplesauce/SauceSession.java  -2
         

Clones added
============
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
- java/src/test/java/com/saucelabs/simplesauce/SauceOptionsTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029&bid=15762536)

java/src/test/java/com/saucelabs/simplesauce/SauceOptionsTest.java
```diff
@@ -1,60 +1,244 @@
...
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.*;
```
nadvolod: Codacy Issue found: [Avoid unused imports such as 'org.junit.Assert'](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029)

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,149 @@
...
+        MutableCapabilities w3cCapabilities = getSeleniumCapabilities();
+        MutableCapabilities sauceCapabilities = new MutableCapabilities();
+
+        Arrays.stream(this.getClass().getDeclaredFields()).forEach(field -> {
+            try {
+                String fieldName = field.getName();
+                if (!fieldName.equals("seleniumCapabilities")) {
+                    Object value = getCapability(fieldName);
+                    String key = fieldName.equals("prerunUrl") ? "prerun" : fieldName;
+                    if (value == null) {
```
nadvolod: Codacy Issue found: [Avoid empty if statements](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029)

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,149 @@
...
+    public MutableCapabilities toCapabilities() {
+        MutableCapabilities w3cCapabilities = getSeleniumCapabilities();
+        MutableCapabilities sauceCapabilities = new MutableCapabilities();
+
+        Arrays.stream(this.getClass().getDeclaredFields()).forEach(field -> {
+            try {
+                String fieldName = field.getName();
+                if (!fieldName.equals("seleniumCapabilities")) {
+                    Object value = getCapability(fieldName);
+                    String key = fieldName.equals("prerunUrl") ? "prerun" : fieldName;
```
nadvolod: Codacy Issue found: [Position literals first in String comparisons](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029)

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
```diff
@@ -1,143 +1,222 @@
...
+        assertEquals(Platform.MAC, capabilities.getCapability("platformName"));
         driver.quit();
     }
 
+    // TODO: This needs to get mocked
     @Test
-    @Ignore("Not sure how to make this work with Mockito. To make sure that the .quit() is actually called on the webDriver")
-    public void stop_noParams_callsDriverQuit() {
-        WebDriver mockDriver = mock(WebDriver.class);
+    public void stopsSessionWithPassingResult() throws MalformedURLException {
```
nadvolod: Codacy Issue found: [JUnit tests should include assert() or fail()](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029)

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
```diff
@@ -1,143 +1,222 @@
...
+
+        sauceSession.start();
+        sauceSession.stop(TestResult.PASS);
+    }
 
-        sauce.start();
-        sauce.stop();
+    // TODO: This needs to get mocked
+    @Test
+    public void stopsSessionWithFailingResult() throws MalformedURLException {
```
nadvolod: Codacy Issue found: [JUnit tests should include assert() or fail()](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029)

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,149 @@
...
+    }
+
+    public MutableCapabilities toCapabilities() {
+        MutableCapabilities w3cCapabilities = getSeleniumCapabilities();
+        MutableCapabilities sauceCapabilities = new MutableCapabilities();
+
+        Arrays.stream(this.getClass().getDeclaredFields()).forEach(field -> {
+            try {
+                String fieldName = field.getName();
+                if (!fieldName.equals("seleniumCapabilities")) {
```
nadvolod: Codacy Issue found: [Position literals first in String comparisons](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029)

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,149 @@
...
+                    Object value = getCapability(fieldName);
+                    String key = fieldName.equals("prerunUrl") ? "prerun" : fieldName;
+                    if (value == null) {
+                        // noop
+                    } else if (isInEnum(key, Options.W3C.class)) {
+                        w3cCapabilities.setCapability(fieldName, value);
+                    } else if (isInEnum(key, Options.SAUCE.class)) {
+                        sauceCapabilities.setCapability(fieldName, value);
+                    } else if (key.equals("publicRestricted")) {
```
nadvolod: Codacy Issue found: [Position literals first in String comparisons](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029)

nadvolod:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/SauceDataCenter.java  1
- java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java  6
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
- java/src/main/java/com/saucelabs/simplesauce/TestResult.java  1
         

Complexity decreasing per file
==============================
+ java/src/main/java/com/saucelabs/simplesauce/SauceSession.java  -2
         

Clones added
============
- java/src/test/java/com/saucelabs/simplesauce/SauceOptionsTest.java  1
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029&bid=15762536)

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,149 @@
...
+        MutableCapabilities w3cCapabilities = getSeleniumCapabilities();
+        MutableCapabilities sauceCapabilities = new MutableCapabilities();
+
+        Arrays.stream(this.getClass().getDeclaredFields()).forEach(field -> {
+            try {
+                String fieldName = field.getName();
+                if (!"seleniumCapabilities".equals(fieldName)) {
+                    Object value = getCapability(fieldName);
+                    String key = "prerunUrl".equals(fieldName) ? "prerun" : fieldName;
+                    if (value == null) {
```
nadvolod: Codacy Issue found: [Avoid empty if statements](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029)

titusfortner:
ironically my unit tests are failing when run in parallel. Apparently I'm not properly mocking out the environment variables with the `@Rule` after all.

nadvolod:


```diff

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/SauceDataCenter.java  1
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
- java/src/main/java/com/saucelabs/simplesauce/TestResult.java  1
- java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java  8
         

Complexity decreasing per file
==============================
+ java/src/main/java/com/saucelabs/simplesauce/SauceSession.java  -2
         

Clones added
============
- java/src/test/java/com/saucelabs/simplesauce/SauceOptionsTest.java  1
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029&bid=15762536)

nadvolod:


```diff

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java  8
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
- java/src/main/java/com/saucelabs/simplesauce/TestResult.java  1
- java/src/main/java/com/saucelabs/simplesauce/SauceDataCenter.java  1
         

Complexity decreasing per file
==============================
+ java/src/main/java/com/saucelabs/simplesauce/SauceSession.java  -2
         

Clones added
============
- java/src/test/java/com/saucelabs/simplesauce/SauceOptionsTest.java  1
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029&bid=15762536)

nadvolod:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/TestResult.java  1
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
- java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java  8
- java/src/main/java/com/saucelabs/simplesauce/SauceDataCenter.java  1
         

Complexity decreasing per file
==============================
+ java/src/main/java/com/saucelabs/simplesauce/SauceSession.java  -2
         

Clones added
============
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
- java/src/test/java/com/saucelabs/simplesauce/SauceOptionsTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029&bid=15762536)

java/src/main/java/com/saucelabs/simplesauce/Options.java
```diff
@@ -0,0 +1,18 @@
+package com.saucelabs.simplesauce;
+
+public enum Options {
+    a;
```
shs96c: Surely some mistake?

java/src/main/java/com/saucelabs/simplesauce/Options.java
```diff
@@ -0,0 +1,18 @@
...
+
+    public enum W3C {
+        browserName, browserVersion, platformName, pageLoadStrategy,
```
shs96c: The normal java-esque way of handling enums is to have the things be in SHOUTY_CASE. I guess you're making it so that these can be passed in as capability names? Perhaps a clarify comment, because this looks odd?

shs96c: Nit: one per line and alphabetised, and then when you come to add more, it's a smaller diff.

java/src/main/java/com/saucelabs/simplesauce/Options.java
```diff
@@ -0,0 +1,18 @@
...
+        acceptInsecureCerts, proxy, setWindowRect, timeouts, strictFileInteractability,
+        unhandledPromptBehavior
+    }
+
+    public enum SAUCE {
+        avoidProxy, build, capturePerformance, chromedriverVersion, commandTimeout,
```
shs96c: Same nit about "one per line" here

java/src/main/java/com/saucelabs/simplesauce/SauceDataCenter.java
```diff
@@ -0,0 +1,15 @@
+package com.saucelabs.simplesauce;
+
+import lombok.Getter;
+
+public enum SauceDataCenter {
+    US_WEST("ondemand.us-west-1.saucelabs.com"),
+    US_EAST("ondemand.us-east-1.saucelabs.com"),
+    EU_CENTRAL("ondemand.eu-central-1.saucelabs.com");
+
+    @Getter private final String value;
```
shs96c, line 10: Comment: Lombok is fine, but `ctrl-return` in IJ will allow you to generate this once and for all if you'd like to reduce your dependencies.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
-    public SauceOptions withMac(MacVersion macVersion) {
-        operatingSystem = macVersion.label;
-        browserName = "Safari";
-        return this;
+    // w3c Settings
+    @Getter @Setter private String browserName = "chrome";
+    @Getter @Setter private String browserVersion = "latest";
+    @Getter @Setter private String platformName = "Windows 10";
+    @Getter @Setter private String pageLoadStrategy;
+    @Getter @Setter private Boolean acceptInsecureCerts = null;
```
shs96c, line 24: Having a `Boolean` be `null` will cause weird NPEs when you accidentally unbox it. Prefer the `boolean` type whereever possible.

shs96c: Same applies for other boolean fields.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
+    @Getter @Setter private Boolean strictFileInteractability = null;
+    @Getter @Setter private String unhandledPromptBehavior;
+
+    // Sauce Settings
+    @Getter @Setter private Boolean avoidProxy = null;
+    @Setter private String build;
+    @Getter @Setter private Boolean capturePerformance = null;
+    @Getter @Setter private String chromedriverVersion;
+    @Getter @Setter private Integer commandTimeout = null;
+    @Getter @Setter private Map customData = null;
```
shs96c: `Map<String, Object>`

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
+    @Getter @Setter private Map<String, Integer> timeouts;
+    @Getter @Setter private Boolean strictFileInteractability = null;
+    @Getter @Setter private String unhandledPromptBehavior;
+
+    // Sauce Settings
+    @Getter @Setter private Boolean avoidProxy = null;
+    @Setter private String build;
+    @Getter @Setter private Boolean capturePerformance = null;
+    @Getter @Setter private String chromedriverVersion;
+    @Getter @Setter private Integer commandTimeout = null;
```
shs96c, line 36: Are you avoiding primitive types so that you can have a null value?

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
+    @Getter @Setter private Map customData = null;
+    @Getter @Setter private Boolean extendedDebugging = null;
+    @Getter @Setter private Integer idleTimeout = null;
+    @Getter @Setter private String iedriverVersion;
+    @Getter @Setter private Integer maxDuration = null;
+    @Getter @Setter private String name;
+    @Getter @Setter private String parentTunnel;
+    @Getter @Setter private Map prerun;
```
shs96c: `Map<String, Object>`

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
+    @Getter @Setter private String prerunUrl;
+    @Getter @Setter private Integer priority = null;
+    @Getter @Setter private String publicRestricted;
+    @Getter @Setter private Boolean recordLogs = null;
+    @Getter @Setter private Boolean recordScreenshots = null;
+    @Getter @Setter private Boolean recordVideo = null;
+    @Getter @Setter private String screenResolution;
+    @Getter @Setter private String seleniumVersion;
+    @Getter @Setter private List tags = null;
```
shs96c: `List<String>`

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
+    @Getter @Setter private String timeZone;
+    @Getter @Setter private String tunnelIdentifier;
+    @Getter @Setter private Boolean videoUploadOnPass = null;
+
+    public static <E extends Enum<E>> boolean isInEnum(String value, Class<E> enumClass) {
```
shs96c: This seems like a utility method that people will abuse. Consider reducing to `private`, or at least package level if that's not possible.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
+        return Arrays.stream(enumClass.getEnumConstants()).anyMatch(e -> e.name().equals(value));
+    }
+
+    public SauceOptions() {
+        this(new MutableCapabilities());
+    }
+
+    public SauceOptions(MutableCapabilities capabilities) {
```
shs96c: `Capabilities` would be safer here.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
+        seleniumCapabilities = capabilities;
```
shs96c: And this should therefore be `seleniumCapabilities = new MutableCapabilities(capabilities)`

shs96c: That will also help avoid confusing things like:

```java
var caps = new MutableCapabilities();
caps.setCapability("browserName", "firefox");
var sauceOptions = new SauceOptions(caps);
caps.setCapability("browserName", "chrome");

// at this point, SauceOptions and the Capabilities differ.
```

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
+            // Team City
+        } else if (System.getenv("TEAMCITY_PROJECT_NAME") != null) {
+            return System.getenv("TEAMCITY_PROJECT_NAME") + ": " + System.getenv("BUILD_NUMBER");
+            // Default
+        } else {
+            return "Build Time: " + Long.toString(System.currentTimeMillis());
+        }
+    }
+
+    public void setCapabilities(Map<String, Object> capabilities) {
```
shs96c, line 125: Consider overriding `merge`, which basically does exactly this.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
+
+    // TODO: Consider making this public
+    private void setCapability(String key, Object value) {
+        String setter = "set" + key.substring(0, 1).toUpperCase() + key.substring(1);
+        try {
+            Class<?> type = SauceOptions.class.getDeclaredField(key).getType();
+            Method method = SauceOptions.class.getDeclaredMethod(setter, type);
+            method.invoke(this, value);
+        } catch (NoSuchMethodException | NoSuchFieldException | InvocationTargetException | IllegalAccessException e) {
+            e.printStackTrace();
```
shs96c, line 137: If the thing has failed, then you should fail too. Throwing a meaningful exception here would be helpful to the user.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
-        browserVersion = version;
-        return this;
-    }
-
-    public SauceOptions withIE() {
-        browserName = "IE";
-        browserVersion = "latest";
-        return this;
-    }
+public class SauceOptions {
```
shs96c, line 16: Consider implementing `Capabilities` so that this can be passed directly to a `RemoteWebDriver`:

```java
var options = new SauceOptions();
var driver = new RemoteWebDriver(options);
```

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
+        if (capabilities.getCapability("browserName") != null) {
+            browserName = (String) capabilities.getCapability("browserName");
+        }
+    }
+
+    public MutableCapabilities toCapabilities() {
+        MutableCapabilities w3cCapabilities = getSeleniumCapabilities();
+        MutableCapabilities sauceCapabilities = new MutableCapabilities();
+
+        for (Field field : this.getClass().getDeclaredFields()) {
```
shs96c, line 73: This seems awfully complicated. The neater thing to do would be to have `setCapability` check the capability name and set it either on the regular capabilities, or on the sauce capabilities. That means you'll end up with something like:

```java
public SauceOptions {
   private MutableCapabilities topLevel = new MutableCapabilities();
   private MutableCapabilities sauce = new MutableCapabilities();

   public SauceOptions() {
     // Tie the two sets of capabilities together
     topLevel.setCapabilities("sauce:options", sauce);
   }

  ...

  private SauceOptions setCapability(String capabilityName, Object value) {
    Objects.requireNonNull(capabilityName, "Name must be set.");
    if ("sauce:options".equals(capabilityName)) {
      throw new WebDriverException("Cowardly refusing to figure out what to do here");
    }
    if (isInEnum(key, Options.SAUCE.class)) {
      sauce.setCapability(capabilityName, value);
    } else {
      topLevel.setCapability(capabilityName, value);
    }
    return this;
  }
}
```

shs96c: Of course, that's if Lombok can generate a custom setter that calls `setCapability`

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
-    }
+public class SauceOptions {
+    @Getter private MutableCapabilities seleniumCapabilities;
 
-    public SauceOptions withMac(MacVersion macVersion) {
-        operatingSystem = macVersion.label;
-        browserName = "Safari";
-        return this;
+    // w3c Settings
+    @Getter @Setter private String browserName = "chrome";
```
shs96c, line 20: I think Lombok has an option to make setters return `this`, which makes chaining easier. Consider using that?

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
+            return System.getenv("CIRCLE_JOB") + ": " + System.getenv("CIRCLE_BUILD_NUM");
+            // Gitlab
+        } else if (System.getenv("CI") != null) {
+            return System.getenv("CI_JOB_NAME") + ": " + System.getenv("CI_JOB_ID");
+            // Team City
+        } else if (System.getenv("TEAMCITY_PROJECT_NAME") != null) {
+            return System.getenv("TEAMCITY_PROJECT_NAME") + ": " + System.getenv("BUILD_NUMBER");
+            // Default
+        } else {
+            return "Build Time: " + Long.toString(System.currentTimeMillis());
```
shs96c: String concatenation will make coerce the `long` to a `String` anyway.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,150 @@
...
+            Class<?> type = SauceOptions.class.getDeclaredField(key).getType();
+            Method method = SauceOptions.class.getDeclaredMethod(setter, type);
+            method.invoke(this, value);
+        } catch (NoSuchMethodException | NoSuchFieldException | InvocationTargetException | IllegalAccessException e) {
+            e.printStackTrace();
+        }
+    }
+
+    // TODO: Consider making this public
+    private Object getCapability(String capability) throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
```
shs96c, line 142: Prefer catching and handling the exceptions in this method, rather than leaking them out if you're thinking of making this `public`

java/src/main/java/com/saucelabs/simplesauce/SauceSession.java
```diff
@@ -1,153 +1,57 @@
...
-        if (timeouts.getCommandTimeout() != 0) {
-            mutableCapabilities.setCapability("commandTimeout", timeouts.getCommandTimeout());
-        }
-        if (timeouts.getIdleTimeout() != 0) {
-            mutableCapabilities.setCapability("idleTimeout", timeouts.getIdleTimeout());
-        }
-        if (timeouts.getMaxTestDurationTimeout() != 0) {
-            mutableCapabilities.setCapability("maxDuration", timeouts.getMaxTestDurationTimeout());
+    public WebDriver start() throws MalformedURLException {
+        if(sauceUsername == null) {
```
shs96c: Run IJ's code formatter over this. The spacing is odd.

java/src/main/java/com/saucelabs/simplesauce/SauceSession.java
```diff
@@ -1,153 +1,57 @@
...
-
-    //TODO 2 same variables being used differently
-    private MutableCapabilities mutableCapabilities;
-    @Getter private MutableCapabilities currentSessionCapabilities;
-    @Getter private final SauceRemoteDriver sauceDriver;
+    @Getter @Setter private SauceDataCenter sauceDataCenter = SauceDataCenter.US_WEST;
+    @Setter private String sauceUsername = System.getenv("SAUCE_USERNAME");
+    @Setter private String sauceAccessKey = System.getenv("SAUCE_ACCESS_KEY");
+    @Getter private SauceOptions sauceOptions;
     @Getter private WebDriver webDriver;
```
shs96c: Boringly, the canonical field name is `driver` :)

nadvolod (changes requested):
A lot of the stuff makes sense, especially the new Sauce API capabilities as they were previously missing. I think that removing all the static typing was a mistake and makes the API less user friendly. Now, the users will continue to have to go to the Sauce documentation to figure out which strings to pass in to make the code work. As opposed to letting the IDE guide them. Furthermore, they will be able to pass in invalid strings that will not work and will make it harder to get started with Sauce Labs. Also, doesn't really make sense to replace everything that's been previously written, reviewed, approved, and in master. I'd request a mixture of the two approaches. Leaving in all the previous functionality and adding on new functionality. With the appropriate refactoring of course.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
joshmgrant: We did discuss this idea, or at least subclassing `SauceOptions` as a `MutableCapabilities`. I don't recall why we didn't do this but there was a good reason.

joshmgrant (approved):
I like the separation of `SauceOptions` and `SauceSession` overall. I'll let the Java professionals take over on the details

nadvolod:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/SauceDataCenter.java  1
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
- java/src/main/java/com/saucelabs/simplesauce/TestResult.java  1
- java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java  8
         

Complexity decreasing per file
==============================
+ java/src/main/java/com/saucelabs/simplesauce/SauceSession.java  -2
         

Clones added
============
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
- java/src/test/java/com/saucelabs/simplesauce/SauceOptionsTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029&bid=15762536)

java/src/main/java/com/saucelabs/simplesauce/SauceDataCenter.java
titusfortner: The goal is less boilerplate / smaller class / easier to read code, not specifically less writing

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
titusfortner: The reasons for this...
The lombok getter for a `boolean` type is `is` instead of `get`, and handling that with reflection was an extra hassle. Additionally, the default value for a `boolean` is `false`, which would override the actual desired default based on w3c, which I do not want to do. If it is `null`, I can just skip it when creating `toCapabilities`.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
titusfortner: yes, exactly, I don't want to override the default value with `0`.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
titusfortner: Oh nice, hadn't considered this usage. Thanks!

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
titusfortner: I don't really understand the appeal of the first over the second, and there are several additional complexities that come from relying on a fluent pattern. I get that it is a common pattern, I'm just reluctant to encourage it without a reason.
```java
sauceOptions.setBrowserName("chrome").
                         setBrowserVersion("77").
                         setPlatformName("Windows 10)
```
```java
sauceOptions.setBrowserName("chrome")
sauceOptions.setBrowserVersion("77")
sauceOptions.setPlatformName("Windows 10)
```

java/src/main/java/com/saucelabs/simplesauce/Options.java
titusfortner: This... is an excellent indication that I am using the wrong data structure for this. Changing it to an immutable list.

java/src/main/java/com/saucelabs/simplesauce/Options.java
titusfortner: was needed to compile, again an indication I'm using the wrong thing.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -1,134 +1,147 @@
...
-
-    public SauceOptions withEdge() {
-        browserName = "Edge";
-        browserVersion = "18.17763";
-        return this;
-    }
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Arrays;
```
nadvolod, line 12: Codacy Issue found: [Unused import - java.util.Arrays.](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029)

nadvolod:


```diff

Issues
======
+ Solved 1
- Added 2
           

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/TestResult.java  1
- java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java  8
- java/src/main/java/com/saucelabs/simplesauce/SauceDataCenter.java  1
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
         

Complexity decreasing per file
==============================
+ java/src/main/java/com/saucelabs/simplesauce/SauceSession.java  -2
         

Clones added
============
- java/src/test/java/com/saucelabs/simplesauce/SauceOptionsTest.java  1
- java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029&bid=15762536)

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
nadvolod, line 12: Codacy Issue found: [Avoid unused imports such as 'java.util.Arrays'](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4721029)

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
titusfortner: Yes, I would definitely use this if we change SauceOptions to implement Capabilities instead of generating an instance of one.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
titusfortner: My current reasons for not doing this (and I might spike out an implementation to verify):
1. SauceOptions is designed to be consumed by SauceSession (for "reasons") not RemoteWebDriver, so it would be confusing to create an instance that "shouldn't" be used in other places
2. I'd like to be strict about what fields are allowed in this class (e.g. not supporting a generic `setCapability`)
3. I don't think it would be compatible with lombok, which I'm not tied to, but it would mean hard coding all of the setter/getters or abandoning point 2.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
titusfortner: Yeah, if we're using lombok, I wouldn't want to create a custom setter. It's there to remove boilerplate, and as soon as it is no longer boilerplate, then imo it shouldn't be used (for reasons of explicitness which we agree on). :)

I'd rather not write custom accessors for everything, and I think a generic `setCapability` introduces other problems. So, I'm reluctant to start with either of these, but I'll evaluate if I can accomplish the other things I want without it.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
nadvolod: @titusfortner I'm not recommending either of those approaches. The cleaner, simpler API approach is:
```
sauceOptions.withChrome("77").withWindows10();
```
One opportunity of messing up something here because obviously someone can pass in an unsupported Chrome version, but a risk we're willing to accept.

**Versus:**
```
sauceOptions.setBrowserName("chrome")
sauceOptions.setBrowserVersion("77")
sauceOptions.setPlatformName("Windows 10)
```
Three opportunities of messing something up here because each of the strings can be either fat fingered or incorrect.
And this is a simple example, adding even more Sauce Options continues to decrease the potential success of a user to get the `sauceOptions` object correct.

titusfortner:
Code from this PR is going to be broken into smaller chunks for separate consideration.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
shs96c: You can avoid the `is` prefix and just get straight getters using the Lombok `lombok.getter.noIsPrefix` configuration option. I'd strongly suggest using `java.beans.Introspector` as the entry-point for your reflection-based magic: that thing's designed for handling "properties" of beans nicely.

java/pom.xml
```diff
@@ -4,6 +4,11 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
+    <properties>
+        <maven.compiler.target>9</maven.compiler.target>
```
shs96c, line 8: I suspect a large number of corporate users are still targeting java 8, and haven't yet made the jump to Java 9. I'd love to be wrong.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
shs96c: Fair enough. If you're never expecting to pass this to a `RemoteWebDriver`, then things are a little different.

shs96c (commented):
Just chatted with @titusfortner, and it sounds like this is being reworked. I'll stop the review now :)

PR #93, opened 2019-12-24
[java] Mock environment variables with PowerMockito

titusfortner:
This PR removes the need to have a constructor for Managing Environment.
All previous tests passing, though the environment variable tests were moved into the Session Test file now.

I had to remove the data provided since there is only one `@RunWith` allowed, so I just copy/pasted into 6 standalone tests.

PR #94, opened 2019-12-25
[java] set username and access key in URL not capabilities

titusfortner:
1. Since we are encapsulating this we need to avoid using capabilities for security reasons
2. Stores URL as URL type
3. Allows direct setting of URL for companies like Schwab that don't go to Sauce directly

Doing the validation of username and accesskey does not belong in a getter, and needs to be moved to start method, but that will be easier after #93


nadvolod:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4730692&bid=15794809)

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
```diff
@@ -43,15 +46,31 @@ public void sauceSession_defaultSauceOptions_returnsChromeBrowser() {
...
-        assertEquals(expectedDataCenterUrl, sauce.getSauceDataCenter());
+        String expectedDataCenterEndpoint = DataCenter.US_WEST.getEndpoint();
+        assertEquals(expectedDataCenterEndpoint, sauce.getSauceDataCenter());
+    }
+
+    @Test
+    public void defaultSauceURL() throws MalformedURLException {
+        String dataCenterEndpoint = DataCenter.US_WEST.getEndpoint();
+        String user = System.getenv("SAUCE_USERNAME");
+        String key = System.getenv("SAUCE_ACCESS_KEY");;
```
nadvolod: Codacy Issue found: [An empty statement (semicolon) not part of a loop](https://app.codacy.com/app/SauceLabs/simple_sauce/pullRequest?prid=4730692)

java/src/main/java/com/saucelabs/simplesauce/SauceSession.java
```diff
@@ -118,9 +119,18 @@ private MutableCapabilities setRemoteDriverCapabilities(MutableCapabilities sauc
         return currentSessionCapabilities;
     }
 
-    private void tryToCreateRemoteWebDriver(String sauceLabsUrl) {
+    public URL getSauceUrl() throws MalformedURLException {
+        if (sauceUrl != null) {
+            return sauceUrl;
+        } else {
+            String url = "https://" + getUserName() + ":" + getAccessKey() + "@" + sauceDataCenter + ":443/wd/hub";
```
mmerrell: Shouldn't need the 443 here

java/src/test/java/com/saucelabs/simplesauce/DataCenterTest.java
```diff
@@ -9,19 +9,19 @@
...
     {
-        assertEquals("https://ondemand.us-east-1.saucelabs.com/wd/hub",
+        assertEquals("ondemand.us-east-1.saucelabs.com",
                 DataCenter.US_EAST.getEndpoint());
     }
     @Test
     public void euCentralDataCenterUrl_isCorrect()
     {
-        assertEquals("https://ondemand.eu-central-1.saucelabs.com/wd/hub",
+        assertEquals("ondemand.eu-central-1.saucelabs.com",
```
mmerrell: This feels like a weird way to test a string constant... I'm not 100% opposed to it, but it feels like a potential point of maintenance

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
```diff
@@ -43,15 +46,31 @@ public void sauceSession_defaultSauceOptions_returnsChromeBrowser() {
...
     @Test
     public void startSession_defaultConfig_usWestDataCenter() {
-        String expectedDataCenterUrl = DataCenter.US_WEST.getEndpoint();
-        assertEquals(expectedDataCenterUrl, sauce.getSauceDataCenter());
+        String expectedDataCenterEndpoint = DataCenter.US_WEST.getEndpoint();
+        assertEquals(expectedDataCenterEndpoint, sauce.getSauceDataCenter());
+    }
+
+    @Test
+    public void defaultSauceURL() throws MalformedURLException {
```
mmerrell: I haven't implemented this yet, so I don't know--are people going to have to try/catch for MalformedURLException all over their code?

mmerrell (commented):
My concerns are very minor, but I'd like to see if we can minimize the "throws MalformedURLException" in the interface

java/src/main/java/com/saucelabs/simplesauce/SauceDriverImpl.java
```diff
@@ -9,10 +9,8 @@
 public class SauceDriverImpl implements SauceRemoteDriver {
     private RemoteWebDriver remoteDriver;
 
-    public RemoteWebDriver createRemoteWebDriver(String seleniumServer, MutableCapabilities capabilities)
-        throws MalformedURLException
-    {
-        remoteDriver = new RemoteWebDriver(new URL(seleniumServer), capabilities);
+    public RemoteWebDriver createRemoteWebDriver(URL seleniumServer, MutableCapabilities capabilities) {
```
nadvolod, line 11: This will cause problems down below where @mmerrell and myself have concerns. I'd like to make sure that we're catching this exception and not forcing users to do `public void test() throws MalformedURLException ` for every test.

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
nadvolod: I'm with @mmerrell . Out of all the great things SimpleSauce does, I was the happiest to never have a single test do `defaultSauceURL() throws MalformedURLException`. It made me 😃 . I'd request to go back to that state

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
```diff
@@ -43,15 +46,31 @@ public void sauceSession_defaultSauceOptions_returnsChromeBrowser() {
...
+        String dataCenterEndpoint = DataCenter.US_WEST.getEndpoint();
+        String user = System.getenv("SAUCE_USERNAME");
+        String key = System.getenv("SAUCE_ACCESS_KEY");
+        URL expetedSauceUrl = new URL("https://" + user + ":" + key + "@" + dataCenterEndpoint + ":443/wd/hub");
+        assertEquals(expetedSauceUrl, sauce.getSauceUrl());
+    }
+
+    @Test
+    public void setsSauceURLDirectly() throws MalformedURLException {
+        sauce.setSauceUrl(new URL("http://example.com"));
```
nadvolod, line 68: I have some confusion here. Why would someone set a "Sauce URL", is that really what a user cares about? Or does the user care about setting the data center?
`sauce.setSauceDataCenter(DataCenter.UsWest)`
Regardless of the approach, I'd want an enum holding those values and avoid potential errors from strings.

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
```diff
@@ -43,15 +46,31 @@ public void sauceSession_defaultSauceOptions_returnsChromeBrowser() {
...
-        String expectedDataCenterUrl = DataCenter.US_WEST.getEndpoint();
-        assertEquals(expectedDataCenterUrl, sauce.getSauceDataCenter());
+        String expectedDataCenterEndpoint = DataCenter.US_WEST.getEndpoint();
+        assertEquals(expectedDataCenterEndpoint, sauce.getSauceDataCenter());
+    }
+
+    @Test
+    public void defaultSauceURL() throws MalformedURLException {
+        String dataCenterEndpoint = DataCenter.US_WEST.getEndpoint();
+        String user = System.getenv("SAUCE_USERNAME");
```
nadvolod: This test will invalidly fail if the Env Variable doesn't exist on the system

nadvolod (commented):
Left a few comments about some potential issues

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
titusfortner: Again, I'm not an expert on managing exceptions, but I'm thinking:

Bindings handles `MalformedURLException` for this:
```
session.start();
```
User handles this `MalformedURLException` for this:
```
session.getSauceUrl();
```

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
titusfortner: This is an easy answer. Schwab needed to set the URL. They had a relay and it needed to be overridden. :)

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
titusfortner: No it is handled here: https://github.com/saucelabs/simple_sauce/blob/8cd6786b0d1e10fdef2de2dfe162e43bee15d4a0/java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java#L30-L31

Though, this way of testing gets to go away with #97

titusfortner:
@mmerrell & @nadvolod I think there is a misunderstanding with which methods rescue which Malformed URL exceptions.

It only is an issue if user calls the public getter `session.getSauceUrl`. I guess we can handle that one as well, but it feels like if we are using it effectively as a private method in our start code, we should handle it, but if a user is asking the question directly, they should handle it.

I'm totally ok with changing it if that's the consensus.

titusfortner:
ok, decided I don't care enough; updated code handles all Malformed URL exceptions.

@nadvolod let me know if I missed any of your feedback; I think latest is good. Should be able to use the "Squash and Commit" button so only the first commit name is used for all the changes in one commit.

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
nadvolod: Do you agree that's more of an exception than the rule? Would it make sense to have 2 methods then? `setSauceUrl` and `setSauceDataCenter`? The first for Shwab, the latter for 99% of other users?

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
nadvolod: I don't believe the code you showed me has any control over this line of code, right?
`System.getenv("SAUCE_USERNAME");` that's a call to the actual OS API.
But, if it's going away, I guess this conversation is moot...

java/src/test/java/com/saucelabs/simplesauce/SauceSessionTest.java
titusfortner: hmm, I'm not sure what your concern is here, this is what we have right now. Schwab can `setSauceUrl` which we don't recommend. Everyone else sets username, access key & data center.

nadvolod: So there will the be capability to `setDataCenter()` somewhere? Guessing on `SauceOptions` ? So that someone with a normal use case doesn't have to pass a URL string? @titusfortner

titusfortner: Oh, good call. Let's add a @Setter to that field, I thought it was already there. Which... brings me to another point in that the field should be the enum for the purposes of the setter... grrr, let me add another commit.

PR #95, opened 2019-12-25
[java] remove driver constructor and driver wrapper

titusfortner:
Driver doesn't belong as a constructor to Session class.
This PR creates the mocks necessary to test Session class without passing it into the Session.

Additionally, it returns a RemoteWebDriver without wrapping it in a SauceWebDriver

PR #96, opened 2019-12-25
[java] Separate the Concerns between Session and Options

titusfortner:
This is based on #93 #94 and #95 

It delegates all options related things to Options class


titusfortner:
This needs reworking. Let's agree on the other PRs before continuing with the new things.

PR #104, opened 2019-12-28
[java] Make with methods more consistent

titusfortner:
Added EdgeVersion enums, not sure why it wasn't used before?
I didn't update SafariVersion, but that also seems weird to only have one value, does it need more?

We do not want to set w3c = true for any browser that doesn't default to it. We might want to error if Chrome version is less than 75.

I added two tests that need to be able to pass and currently do not. I removed the defaults that prevented them from passing. I removed the tests that tested those defaults.

If we're using enum it does not make sense to have the user convert it to a String.
In fact, I don't think we should allow String values if we have enums. 

Browsers need to be named the same way as in the platform configurator; case matters.

That brings me to the next bit, which is I think everything that can be managed as an enum instead of a String should do that. I have a couple other PRs to show that off soon.


nadvolod:


```diff

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/EdgeVersion.java  1
         

Complexity decreasing per file
==============================
+ java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/SauceLabs/simple_sauce/pullRequest?prid=4740738&bid=15822722)

nadvolod:
@titusfortner just overall comments on your concerns above

> Added EdgeVersion enums, not sure why it wasn't used before?
Just a result of Agile development

> I didn't update SafariVersion, but that also seems weird to only have one value, does it need more?


We do not want to set w3c = true for any browser that doesn't default to it. We might want to error if Chrome version is less than 75.

I added two tests that need to be able to pass and currently do not. I removed the defaults that prevented them from passing. I removed the tests that tested those defaults.

If we're using enum it does not make sense to have the user convert it to a String.
In fact, I don't think we should allow String values if we have enums.

Browsers need to be named the same way as in the platform configurator; case matters.

That brings me to the next bit, which is I think everything that can be managed as an enum instead of a String should do that. I have a couple other PRs to show that off soon

nadvolod:


```diff

Complexity increasing per file
==============================
- java/src/main/java/com/saucelabs/simplesauce/EdgeVersion.java  1
         

Complexity decreasing per file
==============================
+ java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/SauceLabs/simple_sauce/pullRequest?prid=4740738&bid=15822722)

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -3,38 +3,26 @@
...
         browserName = BrowserType.CHROME;
         return this;
     }
-    public SauceOptions withSafari()
-    {
-        return withMac(MacVersion.Mojave);
-    }
 
-    public SauceOptions withSafari(SafariVersion version) {
+    public SauceOptions withSafari() {
```
nadvolod, line 18: This is interesting, how will it know which Mac version to use?

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -62,73 +50,56 @@ public SauceOptions withWindows7() {
...
     }
 
-    //TODO notice the duplication below with edge.
-    //Maybe could be moved to a separate class so we can do withEdge().16_16299();
-    //Or withEdge().version(EdgeVersion.16_16299);
     public SauceOptions withEdge18() {
-        withEdge();
-        browserVersion = "18.17763";
-        return this;
+        return withEdge(EdgeVersion._18);
```
nadvolod, line 57: Nice, solid update here 👍

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
```diff
@@ -62,73 +50,56 @@ public SauceOptions withWindows7() {
...
+    public SauceOptions withFirefox() {
+        browserName = "firefox";
         return this;
     }
 
-    public SauceOptions withIE(String version) {
-        browserName = "IE";
-        browserVersion = version;
+    public SauceOptions withIE(IEVersion version) {
+        browserName = "internet explorer";
```
nadvolod, line 91: Wow, will this work as it used to be "IE"?

java/src/test/java/com/saucelabs/simplesauce/MacOsTest.java
```diff
@@ -47,15 +47,6 @@ public void defaultSafari_browserVersionIs12_0() {
         assertEquals("latest", safariVersionSetThroughSauceSession);
     }
 
-    @Test
-    public void defaultSafari_macOsVersionIsMojave() {
```
nadvolod, line 51: Why did you choose to remove this test?

nadvolod (commented):
> I didn't update SafariVersion, but that also seems weird to only have one value, does it need more?

Yes, it would. Just something that got left in the dust.

> In fact, I don't think we should allow String values if we have enums.

Agreed, if that was happening anywhere, it was likely an oversight

> Browsers need to be named the same way as in the platform configurator

I noticed a bunch of the updates here. Maybe it makes sense to store these strings in a single constants class and just test that class for the correct values. Then just use that class everywhere so that we don't mess up the strings?

java/src/test/java/com/saucelabs/simplesauce/MacOsTest.java
titusfortner: The test is removed because the functionality is removed because the code needs to be able to pass these tests: https://github.com/saucelabs/simple_sauce/pull/104/files#diff-6a67e96d9a405e84d935eea01fd071efR87-R103

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
titusfortner: Honestly, I don't know how "IE" works, I don't think it should.

java/src/main/java/com/saucelabs/simplesauce/SauceOptions.java
titusfortner: The user needs to set that separately.

titusfortner:
> Maybe it makes sense to store these strings in a single constants class and just test that class for the correct values

Yes, this is what I did in the other PRs, it can all get rolled together.

titusfortner:
closing this for now

PR #105, opened 2019-12-28
Selenium Approach to Accessors

titusfortner:
Regardless of the With Statements, we need a baseline approach to setting values.
We should either follow the NetBeans Approach (#106) or the Selenium Approach. This is the Selenium Approach.

(This is getting compared to a branch with no accessors just to see an apples-to-apples comparison)

Pro: 
* Enums (when used) protect users from incorrect values
* List of allowed capabilities allows us to throw an exception if something is not recognized
* Easy to toggle fluent support
* More Straightforward implementation 

Con:
* No Type checks on capabilities; no way to enforce usage of enums
* Values set as Strings get returned as enums
* No easy to see list of what capabilities are allowed

This is what it looks like. Note that we can have these be fluent or not. You know my preference, but it's not the hill I want to die on. :)

```java
// Setting values
FirefoxOptions firefoxOptions = new FirefoxOptions();
firefoxOptions.addArguments("--foo");
firefoxOptions.addPreference("foo", "bar");

sauceOptions = new SauceOptions(firefoxOptions);

// setCapability() takes String / Object / Enum values as necessary
sauceOptions.setCapability("pageLoadStrategy", Options.PageLoadStrategy.EAGER)
            .setCapability("acceptInsecureCerts", true)
            .setCapability("unhandledPromptBehavior", "ignore);

Map<Options.Timeouts, Integer> timeouts = new HashMap<>();
timeouts.put(Options.Timeouts.IMPLICIT, 1);
timeouts.put(Options.Timeouts.PAGE_LOAD, 100);
timeouts.put(Options.Timeouts.SCRIPT, 10);
sauceOptions.setCapability("timeouts", timeouts);

// Getting Values --> Values set as Strings return as 
assertEquals(Options.PageLoadStrategy.EAGER, sauceOptions.getCapability("pageLoadStrategy"));
assertEquals(Options.UnhandledPromptBehavior.IGNORE, sauceOptions.getCapability("unhandledPromptBehavior"));
```

I'm not sure it is the best way to do it, but I namespaced all of the enums under `Options` which I think should make it easier to see what is available? Let me know if that's dumb, I can change it.

* An exception is thrown if the capability is not recognized
* This converts String values to Enum values where possible, though that might be confusing since we aren't type checking anything, and users may want to get/set compare and this will prevent that
* Overall this implementation is more straightforward, but the user-facing API is not as helpful.

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -30,7 +30,7 @@
         <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
-            <version>1.18.2</version>
+            <version>1.18.10</version>
         </dependency>
         <dependency>
             <groupId>org.seleniumhq.selenium</groupId>
@@ -135,6 +135,14 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-scm-plugin</artifactId>
                 <version>1.9.4</version>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>9</source>
+                    <target>9</target>
+                </configuration>
             </plugin>
         </plugins>
     </build>
--- a/java/src/main/java/com/saucelabs/simplesauce/DataCenter.java
+++ b/java/src/main/java/com/saucelabs/simplesauce/DataCenter.java
@@ -2,14 +2,18 @@ package com.saucelabs.simplesauce;
 
 import lombok.Getter;
 
-enum DataCenter {
+public enum DataCenter {
     US_WEST("ondemand.us-west-1.saucelabs.com"),
     US_EAST("ondemand.us-east-1.saucelabs.com"),
     EU_CENTRAL("ondemand.eu-central-1.saucelabs.com");
 
-    @Getter private final String endpoint;
+    @Getter private final String value;
 
-    DataCenter(String endpoint) {
-        this.endpoint = endpoint;
+    DataCenter(String value) {
+        this.value = value;
+    }
+
+    public String toString() {
+        return value;
     }
 }
--- a/java/src/main/java/com/saucelabs/simplesauce/Options.java
+++ b/java/src/main/java/com/saucelabs/simplesauce/Options.java
@@ -1,0 +1,261 @@
+package com.saucelabs.simplesauce;
+
+import lombok.Getter;
+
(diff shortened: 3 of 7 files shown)


nadvolod:
```
FirefoxOptions firefoxOptions = new FirefoxOptions();
firefoxOptions.addArguments("--foo");
firefoxOptions.addPreference("foo", "bar");
sauceOptions = new SauceOptions(firefoxOptions);
```
^^^ This part makes sense. Clean and simple, just reusing what Selenium gives us.

```
// setCapability() takes String / Object / Enum values as necessary
sauceOptions.setCapability("pageLoadStrategy", Options.PageLoadStrategy.EAGER)
            .setCapability("acceptInsecureCerts", true)
            .setCapability("unhandledPromptBehavior", "ignore);
```
^^^ No matter how hard I try, I see little value in having the ability to pass in strings into `setCapability()`. Extremely error prone and zero guidance form the IDE regarding what's possible. I prefer our users don't rely on our documentation which has never been really good and I have lots of doubts if it ever will be. Rather, I want them to rely on us to provide a clean API that they can navigate in their IDE.

nadvolod (commented):
```
FirefoxOptions firefoxOptions = new FirefoxOptions();
firefoxOptions.addArguments("--foo");
firefoxOptions.addPreference("foo", "bar");
sauceOptions = new SauceOptions(firefoxOptions);
```
^^^ This part makes sense. Clean and simple, just reusing what Selenium gives us.
```
// setCapability() takes String / Object / Enum values as necessary
sauceOptions.setCapability("pageLoadStrategy", Options.PageLoadStrategy.EAGER)
            .setCapability("acceptInsecureCerts", true)
            .setCapability("unhandledPromptBehavior", "ignore);
```
^^^ No matter how hard I try, I see little value in having the ability to pass in strings into `setCapability()`. Extremely error prone and zero guidance form the IDE regarding what's possible. I prefer our users don't rely on our documentation which has never been really good and I have lots of doubts if it ever will be. Rather, I want them to rely on us to provide a clean API that they can navigate in their IDE.

titusfortner:
Yeah, compared to the Java Beans approach there is a lot less protection for users. If we go this route, we can't enforce Type and if we can't enforce Type, we'd have to support string values like this.

The more generic use case for when we need to be able to accept String values is for this situation: #109

joshmgrant (changes requested):
Declining this in favour of #106

titusfortner:
We've agreed to do #106 instead of this.

PR #106, opened 2019-12-28
Java Beans Approach to Accessors

titusfortner:
Regardless of the With Statements, we need a baseline approach to setting values.
We should either follow the NetBeans Approach or the Selenium Approach (#105). This is the Java Beans Approach.

(This is getting compared to a branch with no accessors just to see an apples-to-apples comparison)

Pro: 
* Enums protect users from incorrect values
* Explicit Fields allow us to ensure types for the various capabilities (and let users see)
* Explicit getter/setters allow the user to know exactly what is available from IDE
* Easy to toggle fluent support

Con:
* Reflection is complicated, then lots of brute force coding for the edge cases
* Converting Strings to enums is complicated

This is what it looks like. Note that we can have these be fluent or not. You know my preference, but it's not the hill I want to die on. :)

```java
FirefoxOptions firefoxOptions = new FirefoxOptions();
firefoxOptions.addArguments("--foo");
firefoxOptions.addPreference("foo", "bar");
sauceOptions = new SauceOptions(firefoxOptions);

// Setting Values
sauceOptions.setBuild("CUSTOM BUILD: 12")
            .setPageLoadStrategy(Options.PageLoadStrategy.EAGER)
            .setUnhandledPromptBehavior(Options.UnhandledPromptBehavior.ACCEPT)
            .setAcceptInsecureCerts(true);

Map<Options.Timeouts, Integer> timeouts = new HashMap<>();
timeouts.put(Options.Timeouts.IMPLICIT, 1);
timeouts.put(Options.Timeouts.PAGE_LOAD, 100);
timeouts.put(Options.Timeouts.SCRIPT, 10);
sauceOptions.setTimeouts(timeouts);

// Getting Values --> Values set as Strings return as 
assertEquals(Options.PageLoadStrategy.EAGER, sauceOptions.getPageLoadStrategy());
assertEquals(Options.UnhandledPromptBehavior.IGNORE, sauceOptions.getUnhandledPromptBehavior());
```

I'm not sure it is the best way to do it, but I namespaced all of the enums under `Options` which I think should make it easier to see what is available? Let me know if that's dumb, I can change it.

* This implementation uses reflection to generate the capabilities (`toCapabilities()`)
* The setters can only accept enums if one is defined for it, (no Strings!).
* `setCapability()` is not public, so users are required to use the named methods, or to pass in a `HashMap` instance to `setCapabilities()`. 
* The use case for `setCapabilities()` is to be able to convert serialized data into a HashMap with a lookup instead of a conditional, then pass it into the `SauceOptions` instance. Since the values have to be from serialized data, they can't be enums. There are a number of ugly conditionals here to auto-translate the String value into an appropriate enum value.
* Overall this implementation is less straightforward, but much more helpful to the end user.

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -30,7 +30,7 @@
         <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
-            <version>1.18.2</version>
+            <version>1.18.10</version>
         </dependency>
         <dependency>
             <groupId>org.seleniumhq.selenium</groupId>
@@ -135,6 +135,14 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-scm-plugin</artifactId>
                 <version>1.9.4</version>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>9</source>
+                    <target>9</target>
+                </configuration>
             </plugin>
         </plugins>
     </build>
--- a/java/src/main/java/com/saucelabs/simplesauce/DataCenter.java
+++ b/java/src/main/java/com/saucelabs/simplesauce/DataCenter.java
@@ -2,14 +2,18 @@ package com.saucelabs.simplesauce;
 
 import lombok.Getter;
 
-enum DataCenter {
+public enum DataCenter {
     US_WEST("ondemand.us-west-1.saucelabs.com"),
     US_EAST("ondemand.us-east-1.saucelabs.com"),
     EU_CENTRAL("ondemand.eu-central-1.saucelabs.com");
 
-    @Getter private final String endpoint;
+    @Getter private final String value;
 
-    DataCenter(String endpoint) {
-        this.endpoint = endpoint;
+    DataCenter(String value) {
+        this.value = value;
+    }
+
+    public String toString() {
+        return value;
     }
 }
--- a/java/src/main/java/com/saucelabs/simplesauce/Options.java
+++ b/java/src/main/java/com/saucelabs/simplesauce/Options.java
@@ -1,0 +1,274 @@
+package com.saucelabs.simplesauce;
+
+import lombok.Getter;
+
+import java.util.*;
+
+public class Options {
+    private static final class BrowserLookup {
+        private static final Map<String, String> lookup = new HashMap<String, String>();
+    }
+
+    private static final class PlatformLookup {
+        private static final Map<String, String> lookup = new HashMap<String, String>();
+    }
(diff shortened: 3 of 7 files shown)


nadvolod:
This API is straight 🔥. Personally preferred over https://github.com/saucelabs/simple_sauce/pull/105 because it makes the life of the end user more simple. Making it easier to get started with Sauce and making it harder to veer away from the valley of success.

Although I can see this as a good approach
```
Map<Options.Timeouts, Integer> timeouts = new HashMap<>();
timeouts.put(Options.Timeouts.IMPLICIT, 1);
timeouts.put(Options.Timeouts.PAGE_LOAD, 100);
timeouts.put(Options.Timeouts.SCRIPT, 10);
sauceOptions.setTimeouts(timeouts);
```
I'm wondering why we wouldn't want to do something like this here:
```
sauceOptions.Timeouts.setImplicit(1);
sauceOptions.Timeouts.setPageLoad(1);
sauceOptions.Timeouts.setScript(1);
```
I can be swayed in other direction, I'm just interested in hearing why one might be preferred over the other. I see the latter as slightly easier to use for the end-user:

- 3 LOC vs 5
- More guidance from the IDE regarding what's actually possible

nadvolod (commented):
This API is straight 🔥. Personally preferred over #105 because it makes the life of the end user more simple. Making it easier to get started with Sauce and making it harder to veer away from the valley of success.

Although I can see this as a good approach
```
Map<Options.Timeouts, Integer> timeouts = new HashMap<>();
timeouts.put(Options.Timeouts.IMPLICIT, 1);
timeouts.put(Options.Timeouts.PAGE_LOAD, 100);
timeouts.put(Options.Timeouts.SCRIPT, 10);
sauceOptions.setTimeouts(timeouts);
```
I'm wondering why we wouldn't want to do something like this here:
```
sauceOptions.Timeouts.setImplicit(1);
sauceOptions.Timeouts.setPageLoad(1);
sauceOptions.Timeouts.setScript(1);
```
I can be swayed in other direction, I'm just interested in hearing why one might be preferred over the other. I see the latter as slightly easier to use for the end-user:

- 3 LOC vs 5
- More guidance from the IDE regarding what's actually possible

titusfortner:
@nadvolod yes, I like that shorter code better, I'm just not sure what is needed to get that to work?

joshmgrant:
I'm leaning toward this implementation over #105 @titusfortner. I agree with @nikolay-advolodkin that this is easier and protects/helps the user better. This might be better for most (majority? large majority??) of Sauce users since they likely won't be using most of the these options most of the time. 

Yeah, let's go with this implementation.

joshmgrant (approved):

titusfortner:
Closing in favor of #115

PR #152, opened 2020-04-20
make name of eu data center consistent with URL and across bindings

titusfortner:
Consistent endpoint names


joshmgrant (approved):
Looks ok to me! If all tests pass then I'm happy :)

titusfortner:
Handling this in #154

PR #153, opened 2020-04-20
create integration tests for each supported endpoint

titusfortner:
note that this conflicts with #152 so they have to be rebased before merging.

This sets us up for the mobile stuff that is coming next since there are some significant w3c/jwp things that require integration tests like these.


titusfortner:
some things in here are wrong, so we can just discuss in #154

PR #154, opened 2020-04-20
Final 1 0

titusfortner:
this fixes & combines #152 & #153 
I was going to add some other things, but I think this is sufficient for 1.0


nadvolod (approved):

java/src/main/java/com/saucelabs/saucebindings/SaucePlatform.java
```diff
@@ -7,6 +7,7 @@
 import java.util.Set;
 
 public enum SaucePlatform {
+    LINUX("Linux"),
```
joshmgrant, line 10: We have to do this, don't we @titusfortner

joshmgrant (commented):
Looks good except for the LINUX platform.

java/src/main/java/com/saucelabs/saucebindings/SaucePlatform.java
titusfortner: Yeah, it's required for Headless, unless we want to handle headless differently via Session (which I'm also ok with)

titusfortner: I feel like we either have to add this now or remove US_EAST support from Session.

joshmgrant: Nah let's keep it like this for now.

joshmgrant (approved):

PR #162, opened 2020-04-27
[java] support Options constructor with DesiredCapabilities instance

titusfortner:
After reviewing this code with a user, I realized that people have crazy complicated ways that they have put together DesiredCapabilities. Lots of conditionals, sometimes DataProviders, sometimes pulling from property files. The easiest way to get people to switch to using SauceBindings is to just allow them to pass in what they have. 

To mitigate people's reliance on it, I'm thinking we can add a deprecation notice. But then I think we also need to create some tutorials for "future-proofing your DesiredCapabilities"  and maybe add a link in the deprecation.


titusfortner:
Sigh. I want this to work, but it is not the right solution. We need to force w3c compliance and I don't want to do conversions.

PR #164, opened 2020-04-30
[python] mock unit tests

titusfortner:
I think I did this right. :)


python/tests/unit/test_session.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def test_raises_exception_if_data_center_is_invalid(self):
+        session = SauceSession()
+
+        with pytest.raises(ValueError):
+            session.data_center = 'invalid'
+
+
+class TestStart(object):
+    @pytest.fixture(autouse=True)
```
joshgrantsauce, line 57: Don't need `autouse=True` since you're explicitly using the fixture in the test methods

python/tests/unit/test_session.py
```diff
@@ -0,0 +1,137 @@
...
+class TestURL(object):
+
+    def test_user_can_override_default_url(self):
+        session = SauceSession()
+        session.remote_url = 'https://foo:bar@foobar.com/wd/hub'
+
+        assert session.remote_url == 'https://foo:bar@foobar.com/wd/hub'
+
+
+class TestStop(object):
```
joshgrantsauce, line 105: This looks genius, tbh

joshgrantsauce (changes requested):
Only one small fix (which might not even be needed depending on how pytest-mock works)

titusfortner:
merged as part of #163

PR #168, opened 2020-05-26
Mobile prereqs

titusfortner:
A few things I tidied up before working on mobile things.

Convenience method for headless
Don't need to do extra for JS Executor

Until we get rid of TestObject endpoints (where we pass in a token), we're going to need the authentication things to be in the options rather than the URL.

### DIFF
--- a/java/src/main/java/com/saucelabs/saucebindings/DataCenter.java
+++ b/java/src/main/java/com/saucelabs/saucebindings/DataCenter.java
@@ -3,9 +3,9 @@ package com.saucelabs.saucebindings;
 import lombok.Getter;
 
 public enum DataCenter {
-    US_WEST("ondemand.us-west-1.saucelabs.com"),
-    US_EAST("ondemand.us-east-1.saucelabs.com"),
-    EU_CENTRAL("ondemand.eu-central-1.saucelabs.com");
+    US_WEST("https://ondemand.us-west-1.saucelabs.com/wd/hub"),
+    US_EAST("https://ondemand.us-east-1.saucelabs.com/wd/hub"),
+    EU_CENTRAL("https://ondemand.eu-central-1.saucelabs.com/wd/hub");
 
     @Getter private final String value;
 
--- a/java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
+++ b/java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
@@ -167,7 +167,7 @@ public class SauceOptions {
     }
 
     public MutableCapabilities toCapabilities() {
-        MutableCapabilities sauceCapabilities = new MutableCapabilities();
+        MutableCapabilities sauceCapabilities = addAuthentication();
 
         if (getCapability("jobVisibility") != null) {
             sauceCapabilities.setCapability("public", getCapability("jobVisibility"));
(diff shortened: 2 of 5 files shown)


java/src/main/java/com/saucelabs/saucebindings/DataCenter.java
```diff
@@ -3,9 +3,11 @@
 import lombok.Getter;
 
 public enum DataCenter {
-    US_WEST("ondemand.us-west-1.saucelabs.com"),
-    US_EAST("ondemand.us-east-1.saucelabs.com"),
-    EU_CENTRAL("ondemand.eu-central-1.saucelabs.com");
+    US_LEGACY("https://ondemand.saucelabs.com/wd/hub"),
```
nadvolod: @titusfortner what's the use case for this?

java/src/main/java/com/saucelabs/saucebindings/DataCenter.java
```diff
@@ -3,9 +3,11 @@
...
+    US_WEST("https://ondemand.us-west-1.saucelabs.com/wd/hub"),
+    US_EAST("https://ondemand.us-east-1.saucelabs.com/wd/hub"),
+    HEADLESS("https://ondemand.us-east-1.saucelabs.com/wd/hub"),
```
nadvolod: @titusfortner do we need this, I think it's going to die, right?

java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
```diff
@@ -128,6 +128,12 @@
         knownCITools.put("TeamCity", "TEAMCITY_PROJECT_NAME");
     }
 
+    public static SauceOptions headless() {
```
nadvolod: Nice and easy :) Does it need to be `static`?

java/src/test/java/com/saucelabs/saucebindings/acceptance/SauceDesktopTest.java
```diff
@@ -47,4 +43,16 @@ public void runsEUCentral() {
         assertNotNull(webDriver);
         assertTrue(session.getSauceUrl().toString().contains("eu-central-1"));
     }
+
+    // Same as US_EAST just more descriptive
+    @Test
+    public void createsHeadlessDriver() {
+        SauceOptions sauceOptions = SauceOptions.headless();
+        session = new SauceSession(sauceOptions);
+        session.setDataCenter(DataCenter.HEADLESS);
```
nadvolod: this is a bit weird, can't we set the data center automatically?

java/src/main/java/com/saucelabs/saucebindings/DataCenter.java
titusfortner: We'll see if we need it, but mobile + w3c + us-west/us-east wasn't working, so I was setting that to default.

java/src/main/java/com/saucelabs/saucebindings/DataCenter.java
titusfortner: nope, it's sticking around according to Matt.

java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
titusfortner: to return an instance, yes, that's what the static methods are for.

nadvolod:
@titusfortner all looks preetty good then. I'm just trying to figure out why Codacy stopped running. I think it's when we migrated the repo, that might have broken stuff. I contacted their support

java/src/test/java/com/saucelabs/saucebindings/acceptance/SauceDesktopTest.java
titusfortner: Yeah, it just added an extra complication. You know, let's just pull this one out for now as a "nice to have" in the future

nadvolod (approved):

titusfortner:
@nadvolod I removed the "LEGACY" and "HEADLESS" parts; if tests pass I'll merge this in.

PR #169, opened 2020-05-26
Android browser emusim

titusfortner:
Update: 
1. I changed to code to use `SauceOptions` instead of `MobileSauceOptions`
2. I added static methods to `SauceOptions` for all the browsers + android + ios
3. I added the iOS things
4. Removed browser default of "latest" since it doesn't work with mobile, and Sauce defaults to that now for Desktop
5. Added TestWatcher to Integration Tests
6. This is missing unit tests
7. Integration Tests will not pass as-is, since they rely on a locally built Appium java-client

Ok, I got it working with subclassing and generics:

```java
SauceOptions options = SauceOptions.android();
SauceAndroidSession session = new SauceAndroidSession(options);
AndroidDriver driver = session.start()
```
Then everything else is samesies.

This look like what we want?

still to go:
1. iOS Browser EmuSim
2. Native App EmuSim
3. Real Device Unified Platform Private
4. Real Device Unified Platform Public
5. Real Device  Private TO
6. Real Device Public TO

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -57,8 +57,8 @@
     <dependencies>
         <dependency>
             <groupId>org.mockito</groupId>
-            <artifactId>mockito-all</artifactId>
-            <version>1.10.19</version>
+            <artifactId>mockito-core</artifactId>
+            <version>3.3.3</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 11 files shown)


nadvolod (approved):

titusfortner:
subsumed by #170

PR #170, opened 2020-05-30
[java] support for Browser on EmuSim

titusfortner:
This replaces #169 

Note that this is Draft, because it relies on [a new PR to Appium](https://github.com/appium/java-client/pull/1356) getting merged and released. I've built the jar with that commit locally and used it to ensure that this code works as intended.

The main updates from 169 is that I've added all the tests, and iOS is completely supported

So now it looks like this:
```java
AndroidOptions androidOptions = new AndroidOptions();
SauceOptions sauceOptions = SauceOptions.android(androidOptions);
SauceAndroidSession session = new SauceAndroidSession(sauceOptions);
AndroidDriver driver = session.start()
```

When reviewing, please check out the things I labeled "TODO" for potential considerations.

### DIFF
--- a/java/src/main/java/com/saucelabs/saucebindings/DeviceOrientation.java
+++ b/java/src/main/java/com/saucelabs/saucebindings/DeviceOrientation.java
@@ -1,0 +1,36 @@
+package com.saucelabs.saucebindings;
+
+import lombok.Getter;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public enum DeviceOrientation {
+    LANDSCAPE("landscape"),
+    PORTRAIT("portrait");
+
+    @Getter
+    private final String value;
+
+    private static final class DeviceOrientationLookup {
+        private static final Map<String, String> lookup = new HashMap<String, String>();
+    }
+
+    public static Set keys() {
+        return DeviceOrientationLookup.lookup.keySet();
+    }
+
+    DeviceOrientation(String value) {
+        this.value = value;
+        DeviceOrientationLookup.lookup.put(value, this.name());
+    }
+
+    public static String fromString(String value) {
+        return DeviceOrientationLookup.lookup.get(value);
+    }
+
+    public String toString() {
+        return this.value;
+    }
+}
--- a/java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
+++ b/java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
@@ -1,10 +1,11 @@
 package com.saucelabs.saucebindings;
 
+import io.appium.java_client.android.AndroidOptions;
+import io.appium.java_client.ios.IOSOptions;
+import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.Setter;
-import lombok.AccessLevel;
 import lombok.experimental.Accessors;
-
 import org.openqa.selenium.MutableCapabilities;
 import org.openqa.selenium.Proxy;
 import org.openqa.selenium.chrome.ChromeOptions;
@@ -24,11 +25,12 @@ import java.util.Map;
 @Setter @Getter
 public class SauceOptions {
     @Setter(AccessLevel.NONE) private MutableCapabilities seleniumCapabilities;
+
     public TimeoutStore timeout = new TimeoutStore();
 
     // w3c Settings
     private Browser browserName = Browser.CHROME;
-    private String browserVersion = "latest";
+    private String browserVersion;
     private SaucePlatform platformName = SaucePlatform.WINDOWS_10;
     private PageLoadStrategy pageLoadStrategy;
     private Boolean acceptInsecureCerts = null;
@@ -38,13 +40,27 @@ public class SauceOptions {
     private Boolean strictFileInteractability = null;
     private UnhandledPromptBehavior unhandledPromptBehavior;
 
+    // Defined in Appium
+    // These technically belong in Appium's IOSOptions & AndroidOptions
+    // but Sauce has always had them in its documentation, so it should handle them
+    private String app; // TODO - consider handling URL and auto-add "sauce-storage:"
+    private String appActivity; // Android only
+    private String appPackage; // Android only - TODO - Should this be an Array?
+    private Boolean autoAcceptAlerts; // iOS only
+    private SauceAutomationName automationName;
+    private String deviceName;
+    private DeviceOrientation deviceOrientation; // Similar to Selenium's ScreenOrientation
+    private String platformVersion;
+
     // Sauce Settings
+    private String appiumVersion;
     private Boolean avoidProxy = null;
     private String build;
     private Boolean capturePerformance = null;
     private String chromedriverVersion;
     private Integer commandTimeout = null;
     private Map<String, Object> customData = null;
+    private DeviceType deviceType; // Android Only for generic emulators
     private Boolean extendedDebugging = null;
     private Integer idleTimeout = null;
     private String iedriverVersion;
@@ -66,7 +82,10 @@ public class SauceOptions {
     private Boolean videoUploadOnPass = null;
 
     public static final List<String> primaryEnum = List.of(
+            "automationName",
             "browserName",
+            "deviceOrientation",
+            "deviceType",
             "jobVisibility",
             "pageLoadStrategy",
             "platformName",
@@ -91,13 +110,25 @@ public class SauceOptions {
             "strictFileInteractability",
             "unhandledPromptBehavior");
 
+    public static final List<String> appiumDefinedOptions = List.of(
+            "app",
+            "appActivity",
+            "appPackage",
+            "autoAcceptAlerts",
+            "automationName",
+            "deviceName",
+            "platformVersion",
+            "deviceOrientation");
+
     public static final List<String> sauceDefinedOptions = List.of(
+            "appiumVersion",
             "avoidProxy",
             "build",
             "capturePerformance",
             "chromedriverVersion",
             "commandTimeout",
             "customData",
+            "deviceType",
             "extendedDebugging",
             "idleTimeout",
             "iedriverVersion",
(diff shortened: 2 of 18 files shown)


java/src/main/java/com/saucelabs/saucebindings/DeviceOrientation.java
```diff
@@ -0,0 +1,36 @@
...
+        return DeviceOrientationLookup.lookup.keySet();
+    }
+
+    DeviceOrientation(String value) {
+        this.value = value;
+        DeviceOrientationLookup.lookup.put(value, this.name());
+    }
+
+    public static String fromString(String value) {
+        return DeviceOrientationLookup.lookup.get(value);
```
mykola-mokhnach, line 30: `return this.valueOf(value.toUpperCase())`
or 
`return Stream.of(this.values()).findFirst((x) -> x.value.equalsIgnoreCase(value)).orElse(null)`

java/src/main/java/com/saucelabs/saucebindings/DeviceOrientation.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Getter
+    private final String value;
+
+    private static final class DeviceOrientationLookup {
+        private static final Map<String, String> lookup = new HashMap<String, String>();
+    }
+
+    public static Set keys() {
+        return DeviceOrientationLookup.lookup.keySet();
```
mykola-mokhnach, line 21: Could also be `return Stream.of(this.values()).map((x) -> x.name()).collect(toList())`

java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
```diff
@@ -38,13 +40,27 @@
     private Boolean strictFileInteractability = null;
     private UnhandledPromptBehavior unhandledPromptBehavior;
 
+    // Defined in Appium
+    // These technically belong in Appium's IOSOptions & AndroidOptions
+    // but Sauce has always had them in its documentation, so it should handle them
+    private String app; // TODO - consider handling URL and auto-add "sauce-storage:"
+    private String appActivity; // Android only
```
mykola-mokhnach, line 47: The fact this class includes platform-specific options makes me think it needs to be refactored and split

java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
```diff
@@ -128,6 +159,83 @@
...
+        SauceOptions sauceOptions = new SauceOptions(options);
+        sauceOptions.setPlatformName(SaucePlatform.MAC_MOJAVE);
+        return sauceOptions;
+    }
+
+    public static SauceOptions android(AndroidOptions options) {
+        SauceOptions sauceOptions = new SauceOptions(options);
+        sauceOptions.setPlatformName(SaucePlatform.ANDROID);
+        sauceOptions.setDeviceName("Android GoogleAPI Emulator");
+        sauceOptions.setPlatformVersion("8.1");
```
mykola-mokhnach, line 223: should it be hardcoded?

java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
mykola-mokhnach: to be more specific

```
sauceOptions.setAppActivity
``` 

must not be compilable if `sauceOptions` does not target the Android platform

titusfortner: Yeah, from a CS perspective, this is a "bad class" because not all fields apply to all the Sessions. At the same time, the reason it is bad is because it can theoretically be more difficult to maintain when it isn't abstracted out. The stated goal of the project is to make it easy for our users to get what they want, and when balancing out the added maintenance costs here I believe we should err on the side of usability.

Essentially I don't want users to have to spend time figuring out whether options need to be in `AndroidOptions` or `SauceOptions` when our docs have always included them. If someone needs things that are only specified in Appium docs or tutorials, that's an easier ask to put in a separate class, especially since I expect that to be a small minority of users.

java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
titusfortner: This is a way to set a different default for each driver instead of defining one in the field directly. Honestly, I'd rather Sauce provide a default when it isn't specified, but until then this software should do it. Obviously it can be overridden as is already necessary with current code. 

Also, we probably need a way to track the things that are hard-coded that are likely to change. Ideally we can pull from a config file that can be auto-generated by the API at some point, but that is a lot of work for where we are right now.

java/src/main/java/com/saucelabs/saucebindings/DeviceOrientation.java
titusfortner: Yeah, it took me a while to find an approach that worked for what I needed. This code matches what I'm doing in all the other enums. Is the an advantage to replacing the code in all of them?

java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
titusfortner: I guess I should say that we should be comparing this code to `MutableCapabilities` which has zero protections, and from that standpoint this is better. We could do a better job of evaluating what combinations of capabilities don't work before we send it to Sauce, but I don't think that is MVP for this.

java/src/main/java/com/saucelabs/saucebindings/DeviceOrientation.java
mykola-mokhnach: I would say it makes sense since we get rid of the redundant data structure. All operations could be done using native methods of the Enum class itself

java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
nadvolod: I agree with @mykola-mokhnach here. This causes me a lot of confusion to see a mixture of mobile options and browser based options in a single class. I think that we do want the user to figure out "whether options need to be in AndroidOptions or SauceOptions". `AndroidOption` for Android, `IosOptions` for iOS, `SauceOptions` for VDC. Regardless of our docs, these 3 products have very distinctive responsibility boundaries. I think that should be reflected when a user is trying to use our bindings. Otherwise they'll have this massive class with a massive API. What applies only for android, what applies for iOS? Will also make it difficult for us to name methods. To me it's thee same difference as determining whether to use `AppiumOptions` or `MutableCapabilities`. Yes, as a user we need to know if we're testing web or mobile.

java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
```diff
@@ -289,6 +427,23 @@ private void setEnumCapability(String key, String value) {
                     setBrowserName(Browser.valueOf(Browser.fromString(value)));
                 }
                 break;
+            case "deviceOrientation":
+                if (!DeviceOrientation.keys().contains(value)) {
```
nadvolod, line 431: Wow, this method is getting scary! I created a PR to clean it up https://github.com/saucelabs/sauce_bindings/pull/171

java/src/test/java/com/saucelabs/saucebindings/SauceAndroidSessionTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.openqa.selenium.MutableCapabilities;
+
+import java.net.URL;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.*;
```
nadvolod, line 14: Codacy Issue found: [Avoid unused imports such as 'org.mockito.Mockito'](https://app.codacy.com/manual/SauceLabs/sauce_bindings/pullRequest?prid=5612597)

nadvolod:


```diff

Issues
======
- Added 2
           

Complexity increasing per file
==============================
- java/src/test/java/com/saucelabs/saucebindings/integration/SauceTestWatcher.java  1
- java/src/test/java/com/saucelabs/saucebindings/SauceIOSSessionTest.java  1
- java/src/test/java/com/saucelabs/saucebindings/SauceAndroidSessionTest.java  1
- java/src/main/java/com/saucelabs/saucebindings/DeviceOrientation.java  1
- java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java  9
- java/src/test/java/com/saucelabs/saucebindings/integration/SauceEmuSimBrowserTest.java  1
- java/src/main/java/com/saucelabs/saucebindings/DeviceType.java  1
- java/src/main/java/com/saucelabs/saucebindings/SauceAndroidSession.java  1
- java/src/test/java/com/saucelabs/saucebindings/SauceMobileOptionsTest.java  1
- java/src/main/java/com/saucelabs/saucebindings/SauceIOSSession.java  1
- java/src/main/java/com/saucelabs/saucebindings/SauceAutomationName.java  1
         

Clones added
============
- java/src/test/java/com/saucelabs/saucebindings/SauceMobileOptionsTest.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/SauceLabs/sauce_bindings/pullRequest?prid=5612597&bid=18386551)

java/src/test/java/com/saucelabs/saucebindings/SauceIOSSessionTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.Test;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.openqa.selenium.MutableCapabilities;
+
+import java.net.URL;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.*;
```
nadvolod, line 15: Codacy Issue found: [Avoid unused imports such as 'org.mockito.Mockito'](https://app.codacy.com/manual/SauceLabs/sauce_bindings/pullRequest?prid=5612597)

PR #179, opened 2020-08-05
Clarify JDK version requirements

### LINKED ISSUE #178: Cannot use Java library with Java 8
The prerequisites state that this can be used with a JDK8+ JVM:

https://github.com/saucelabs/sauce_bindings/blob/master/java/README.md#building

But, this jar is compiled with Java 9-only compatibility:

https://github.com/saucelabs/sauce_bindings/blob/da74978d20a7f9cb6e24cc7c6cfb38ec683db76b/java/pom.xml#L172-L179

Thus, when you attempt to add this dependency to another project targeting Java 8 (and specifically with a Java 8 JDK) you get this:

```console
Caused by: java.lang.UnsupportedClassVersionError: com/saucelabs/saucebindings/SauceSession has been compiled by a more recent version of the Java Runtime (class file version 53.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:757)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
```

phillipuniverse:
Resolves #178 as there is intentionally no Java 8 compatibility

### DIFF
--- a/java/README.md
+++ b/java/README.md
@@ -86,7 +86,7 @@ to your `pom.xml` file.
 
 Make sure you have the common Sauce Bindings [prerequisites](https://github.com/saucelabs/sauce_bindings#getting-started-and-prerequisites) set up, as well as
 
--  Java 8+ JDK,
+-  Java 9+ JDK,
 -  Maven,
 -  Your favorite Java IDE ([IntelliJ](https://www.jetbrains.com/idea/download/index.html) preferred but not required).
 


titusfortner:
Oh thanks. We probably need to pull out most of the README and just point people to the actual documentation: https://opensource.saucelabs.com/sauce_bindings/docs/getting-started

PR #183, opened 2020-10-09
[java] support Java 8 by removing Java 9 List.of implementation

titusfortner:
So, it looks like when I refactored at one point I removed most of the functionality that required Java 9, so this should "just work."

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,24 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: JDK 8
+
+on:
+  push:
+    branches: [ master ]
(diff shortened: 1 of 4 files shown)


nadvolod:
@titusfortner that was fantastically easy!! Should we add one Travis pipeline that runs against JDK 8 to make sure stuff works?

nadvolod (commented):
@titusfortner that was fantastically easy!! Should we add one Travis pipeline that runs against JDK 8 to make sure stuff works?

titusfortner:
Sure? I'll see if I can update Travis.

I remember what code I removed. I was using reflection to get all the options class properties dynamically, but then decided I should be explicit and list the ones I needed and just iterate over them.

nadvolod (approved):

mmerrell (approved):
Yeah, this looks good to me. Ship it!

PR #218, opened 2020-10-22
[java] Visual API using Strategy pattern

nadvolod:
This is a cleaner implementation using the Strategy pattern. I personally prefer this approach because it makes the separation of concerns more clear. It also opens up the possibility to follow Open-Closed Principle for other portions of Sauce.
- If we wanted to update the result using an API, we just introduce a new testingStrategy

We would need to update some of the SauceSessionTests to make this work

PR #213, opened 2020-10-21
Visual testing capabilities version 2

nadvolod:
These are visual testing capabilities as we discussed:

Visual testing is enabled like this:
`SauceOptions sauceOptions = new SauceOptions().visual();`

Configuration:
```
sauceOptions.setName("testName");
sauceOptions.setViewportSize("1280x1024");
```

**What I don't like:**
- The user can set visual capabilities without actually running a visual test. Not necessarily bad I think. But is a confusing API. 
In one instance I chose to throw an `OptionForVisualTestingOnlyException` exception if `visualTesting` isn't true. But that's also awkward for a few reasons. Probably want to get away from this, but I included it as an option.
- I don't like adding if/else blocks to the currently working methods to perform different operations if `visualTesting` is enabled. First, that can break the original functionality. Second, it's making the code messy. I think I can resolve this in a cleaner manner using the Strategy pattern as I did [here](https://github.com/saucelabs/sauce_bindings/pull/218).

PR #223, opened 2020-10-30
[java] implement visual options within sauce options

titusfortner:
I broke this into 2 commits. The first is what is completely matching the current structure of the code...
(Almost) all public instance methods in `SauceOptions` are lombok-generated getters & setters

Minimum required
```java
sauceOptions.setVisualProjectName("My Project Name");
```
We need this to work so we can set values from a config file. (see `parsesCapabilitiesFromVisualValues()` test below)

The second commit is optional. It's how I was thinking we might encourage people to use a factory pattern type approach to initializing options with the basic things that make sense, and could be extended to include mobile, etc. It looks like this:
```
SauceOptions sauceOptions = SauceOptions.visual("My Project Name");
```

My current big concern with this is the extent to which the method names for these options make sense external to the context of visual...
for instance:
`projectName` doesn't immediately make me think of visual, so I renamed it "visualProjectName" for the bindings
`ignore` also doesn't really seem obvious. `ignoreCSS` maybe? or `visualIgnore`? 

Makes me wonder if the context  for saying these are Visual and only visual objects might not be worthwhile. But I guess that's not where I think we should go first.

Other Stuff
1. This renames `projectName` to `visualProjectName`
2. This does not rename `ignore` to anything that would be easier to understand
3. This sets Sauce Labs Test Name to Visual Project Name
4. This doesn't allow ignore parameters to be a List, only a String (should get fixed before merged)
5. What should we do when a non-visual test calls takeSnapshot?
6. This should get updated with the Session pattern in #218 
7. @nadvolod I'm not sure why #213 had special logic for [`getScreenerApiKey()`](https://github.com/saucelabs/sauce_bindings/pull/218/files#diff-49a321ee5882d79545dd8f7728516ace51372a046901f91fe3b6dac5532ae440R363)

### DIFF
--- a/java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
+++ b/java/src/main/java/com/saucelabs/saucebindings/SauceOptions.java
@@ -66,6 +66,20 @@ public class SauceOptions {
     private String tunnelIdentifier;
     private Boolean videoUploadOnPass = null;
 
+    // Sauce Visual Settings
+    // https://screener.io/v2/docs/visual-e2e/visual-options
+    private String visualProjectName; // the actual key for this is "projectName"
+    private String viewportSize;
+    private String branch;
+    private String baseBranch;
+    private Map<String, Object> diffOptions = null;
+    private String ignore = null;   // This probably should be a List not a String, but conversions, ugh
+    private Boolean failOnNewStates = null;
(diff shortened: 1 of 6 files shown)


titusfortner:
Superceded by #256

PR #224, opened 2020-10-30
Visual4

titusfortner:
@nadvolod I'm pretty sure this is much more interesting of an approach to you.
This addresses most of the concerns I had in #223, but still a couple outstanding

Features:
* Required to use static method/factory pattern to construct
* Always requires projectName
* Alternate factory constructor takes MutableCapabilities instance as well
* setters and getters use instance method `visual()`
```
SauceOptions sauceOptions = SauceOptions.visual("acceptsChromeOptionsClass", new FirefoxOptions());
sauceOptions.visual().setViewportSize("1900x1200");
```

If we like this, the next question is if we want to make non-backward-compatible changes to how we are differentiating browser, sauce & w3c capabilities? Or just move forward with this approach for mobile and go from there?


titusfortner:
Superceded by #256

PR #226, opened 2020-11-02
Move sauce parameters to be accessed by the sauce() method

titusfortner:
This is what a backward compatible implementation would look like if we treated Sauce parameters the same way we're looking to implement visual parameters in #224 

It's ugly because I don't know how to do dynamic delegation of setters/getters or if that's even possible in Java. :)
This PR also needs to have the new specs added to show accessing sauce parameters the "new" way. (I'll add that in tomorrow)

```
sauceOptions.setBrowserName(Browser.FIREFOX);
sauceOptions.sauce().setBuild("My Build 123");
``` 

I kind of like this approach for Java, but it doesn't really translate well to Python/Ruby, so I'm interested @joshmgrant what you think about this. We don't necessarily have to keep the implementations exactly the same, either.


nadvolod:
This is definitely cleaner and more consistent @titusfortner . Trying to decide if the extra work to make this possible is necessary?

titusfortner:
Hmm, yeah, I'm not sure, honestly...

I'm trying to figure what this looks like big picture... Thinking about it more, it might not be what we want...

So something like this where we distinguish between "w3c, sauce, visual, appium"
```java
sauceOptions.setBrowserName(Browser.FIREFOX);
sauceOptions.sauce().setExtendedDebugging(true);
sauceOptions.visual().setProjectName("Visual Project Name");
```

```java
sauceOptions.setApp(appLocation);
sauceOptions.sauce().setCommandTimeout(60);
sauceOptions.appium().setDeviceName(myDevice);
```

A further idea would be if we could adjust what is available in `Sauce` based on what static method is used to initialize things...

```
SauceOptions sauceOptions = SauceOptions.visual()
sauceOptions.sauce(); // Uses SauceVisualOptions instance that only has Sauce options available for Visual
```

```
SauceOptions sauceOptions = SauceOptions.emusim(androidOptions);
sauceOptions.sauce(); // Uses SauceEmuSimAndroidOptions instance
```

titusfortner:
superseded by #237

PR #227, opened 2020-11-03
Missing visual specs for parallelization + multi-screeenshot

nadvolod:
I'm adding 2 important use cases in visual. One for doing multi screenshots in a single test. One for running a test cross-platform. As we will see, our current implementation of cross-platform doesn't work. I am pretty sure that it's because we by default set a build name which in Screener should not be set.


titusfortner:
Superceded by #256

PR #229, opened 2020-11-09
Capability manager

titusfortner:
This is a continuation of #226 
I didn't care for how SauceLabsOptions subclassed SauceOptions. It doesn't make sense to do `sauceOptions.sauce().setBrowserName(Browser.CHROME)`

This pulls implementations into CapabilitiesManager and EnvironmentManager classes. I'm honestly not sure if this is the right way to do it, but it should be more extensible.

This implementation should be completely 100% backward compatible. Methods that will be removed are marked as deprecated. We need to figure out how to do the right reporting so that explanation shows up in IDE? Not sure how that works, yet.

This PR doesn't add the new "correct" way to set parameters, so this is still in Draft.


titusfortner:
Superseded by combination of #230 & #237

PR #230, opened 2020-11-09
[java] move system and capability commands into their own classes

titusfortner:
I'm going to step through the ideas I discussed today in smaller incremental commits.

@mmerrell if you have time, can you tell me if this is the right way in Java to pull out methods from a class?
I'm creating/storing an instance of CapabilitiesManager in SauceOptions, and then using EnvironmentManager with static methods.

This is ready to merge if everyone is good with it.

### DIFF
--- a/java/src/main/java/com/saucelabs/saucebindings/CapabilityManager.java
+++ b/java/src/main/java/com/saucelabs/saucebindings/CapabilityManager.java
@@ -1,0 +1,106 @@
+package com.saucelabs.saucebindings;
+
+import org.openqa.selenium.MutableCapabilities;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class CapabilityManager {
+    private final SauceOptions options;
+
+    /**
+     * Class constructor created for a specific SauceOptions instance
+     *
+     * @param options the SauceOptions instance using this capabilities manager
+     */
+    public CapabilityManager(SauceOptions options) {
+        this.options = options;
+    }
+
+    /**
+     * Add values of valid capabilities to the capabilities object
+     *
+     * @param capabilities the capabilities instance that valid options get added to
+     * @param validOptions the list of options matching the options being used
+     */
+    public void addCapabilities(MutableCapabilities capabilities, List<String> validOptions) {
+        validOptions.forEach((capability) -> {
+            Object value = getCapability(capability);
+            if (value != null) {
+                capabilities.setCapability(capability, value);
+            }
+        });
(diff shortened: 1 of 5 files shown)


java/src/main/java/com/saucelabs/saucebindings/CapabilityManager.java
```diff
@@ -0,0 +1,45 @@
...
+            Method declaredMethod = null;
+            declaredMethod = options.getClass().getDeclaredMethod(getter);
+            return declaredMethod.invoke(options);
+        } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+            e.printStackTrace();
+            return null;
+        }
+    }
+
+    protected void capabilityValidator(String name, Set values, String value) {
```
nadvolod: @titusfortner would something like `validateCapability()` be a better method name?

nadvolod (commented):
Only requesting one minor change

java/src/main/java/com/saucelabs/saucebindings/CapabilityManager.java
titusfortner: yes, yes it would. Not sure what I had before that this made more sense.

titusfortner:
@nadvolod ok made the update, this should be good.

java/src/main/java/com/saucelabs/saucebindings/SystemManager.java
```diff
@@ -0,0 +1,29 @@
+package com.saucelabs.saucebindings;
+
+public class SystemManager {
+    protected static String get(String key, String errorMessage) {
```
nadvolod: @titusfortner are these static methods going to be a problem when it comes to parallelization?

titusfortner: Not with the way surefire works, but potentially with TestNG, but that's why people need to use thread local. 

That said, it would make sense to test this out with an example suite before getting too far along.

joshmgrant: I'm going to guess this is fine since these methods only return "constant" string values. These values should remain constant across tests/classes anyway.

joshmgrant (approved):
lgtm

java/src/main/java/com/saucelabs/saucebindings/SystemManager.java
titusfortner: Oh, I was looking at wrong code, yes, there is no state needed for these, so no need to instantiate anything.

nadvolod (approved):

titusfortner:
well, I'm not sure we actually need this one now, so hold off on merging it...

titusfortner:
part of #246

PR #231, opened 2020-11-10
[java] move sauce specific options into their own class

titusfortner:
Continuation of #230 so it is comparing to caps_manager branch & in Draft form

* Create Abstract Superclass called "Options"
* Breaks out Sauce Specific Options into a separate class both classes extend Options class
* Creates Deprecated, Backward compatible getters/setters on SauceOptions for supporting until 2.0
* Breaks out Tests into backward compatible / deprecated & recommended API

Only interesting thing here is that fluent pattern works slightly differently
https://github.com/saucelabs/sauce_bindings/compare/caps_manager...subclass_sauce?expand=1#diff-04ebfed2c5f642255616b935927952038b577e44c7a2b216a1edcd05248e063fR162

### DIFF
--- a/java/src/main/java/com/saucelabs/saucebindings/CapabilityManager.java
+++ b/java/src/main/java/com/saucelabs/saucebindings/CapabilityManager.java
@@ -1,0 +1,45 @@
+package com.saucelabs.saucebindings;
+
+import org.openqa.selenium.MutableCapabilities;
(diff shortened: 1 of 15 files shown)


titusfortner:
This doesn't work the way I need to, so I'm removing it; next one will be better...

PR #233, opened 2020-11-11
update Java example code for docs

titusfortner:
Lindsay pointed out that the class names need to be named Test so here are the updates.


nadvolod (approved):
👍

PR #236, opened 2020-11-15
Factory + Builder FTW

titusfortner:
(this is in Draft because it is mostly just making sure that it is the pattern we want to go with; it's comparing with #237 as baseline)

There will be 3 ways to create Sauce Options.

1. The old way (which this PR deprecates)
The idea was to set each option directly onto the `SauceOptions` instance
```
        sauceOptions.setBrowserName("chrome);
        sauceOptions.setAvoidProxy(true);
        sauceOptions.setBuild("Sample Build Name");
```
When looking at additional mobile and visual capabilities, we realized there is too much here to set directly and it can be easily overwhelming.
Tests for the old way are implemented in `SauceOptionsDeprecatedTest.java` file

2. The new way
Everything that goes into `sauce:options` gets accessed via `sauce()`
```
        sauceOptions.setBrowserName("chrome);
        sauceOptions.sauce().setAvoidProxy(true);
        sauceOptions.sauce()setBuild("Sample Build Name");
```
This is a little better, but using this method, there is no way to ensure that only applicable settings are available to the user.
Tests for the new way are implemented in `SauceOptionsTest.java` file

3. The Factory+Builder way!
This makes sure we have a `SauceOptions` instance that only uses valid capabilities for the option selected.

First, use the SauceOptionsFactory to pick the type of test you're running
Then use the build() method to generate the correct options. You need to use sauce() to set sauce specific values like the "new way" above, and you need to use `build()` to generate it all... looks like this:
```
        ChromeOptions chromeOptions = new ChromeOptions();
        chromeOptions.addArgument("--foo");

        SauceOptions sauceOptions = SauceOptionsFactory.chrome(chromeOptions)
                .setPlatformName(SaucePlatform.MAC_CATALINA)
                .sauce()
                        .setBuild("Sample Build Name")
                        .setMaxDuration(300)
                        .build()
                .build();
```
The biggest downside is going to be knowing which factory you want, but theoretically will be easy to tell.
Tests for the Factory+Builder way are implemented in `SauceChromeConfigsTest.java` file

Factory+Builder approach will need to have a lot of code duplication or use the boilerplate, since chaining setters with lombok doesn't work with inheritance. I'm defaulting to more duplication and less boilerplate in this example.

### DIFF
--- a/java/src/main/java/com/saucelabs/saucebindings/BaseOptions.java
+++ b/java/src/main/java/com/saucelabs/saucebindings/BaseOptions.java
@@ -1,0 +1,25 @@
+package com.saucelabs.saucebindings;
+
+import lombok.Getter;
+import org.openqa.selenium.MutableCapabilities;
+
+import java.util.List;
+import java.util.Map;
+
+public abstract class BaseOptions {
+    @Getter protected MutableCapabilities capabilities = new MutableCapabilities();
+    protected CapabilityManager capabilityManager;
+    @Getter public final List<String> validOptions = null;
+
+    // Use Case is pulling serialized information from JSON/YAML, converting it to a HashMap and passing it in
+    // This is a preferred pattern as it avoids conditionals in code
+    public void mergeCapabilities(Map<String, Object> mergingCapabilities) {
+        mergingCapabilities.forEach(this::setCapability);
+    }
+
+    // This dynamically calls setter
+    // Applicable enums must override this method in subclass
+    protected void setCapability(String key, Object value) {
+        capabilityManager.setCapability(key, value);
+    };
+}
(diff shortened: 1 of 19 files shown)


titusfortner:
Another interesting thing we *could do with this pattern.
Advantage: never need to know anything about "SauceOptions" class
Disadvantage: That's a lot of `build()` methods

```
SauceSession.optionsBuilder()
    .chrome()
        .setPlatformName(SaucePlatform.MAC_CATALINA)
        .sauce()
            .setName("My Test Name")
        .build()
    .build()
.build()
```

joshmgrant:
This just looks plain wrong to me 

```SauceSession.optionsBuilder()
    .chrome()
        .setPlatformName(SaucePlatform.MAC_CATALINA)
        .sauce()
            .setName("My Test Name")
        .build()
    .build()
.build()
```

The "advantage" of being to build a specific Sauce session in a "one-liner" might be countered in both increased complexity of this line and also debugging runtime configuration. It's likely there will be valid but "wrong" configurations and this might be difficult to debug if written this way.

titusfortner:
yeah, I agree, we're getting into the clever/complicated territory with that one.

But what about the `build().build()` of it all in the SauceOptions generation?

titusfortner:
implemented in #246

PR #237, opened 2020-11-16
[java] move sauce specific options into their own class

titusfortner:
This is in DRAFT status because it relies on #230 
This supersedes PR #226 

All Sauce Labs options will be accessed with `sauce()` method instead of directly.
This is completely backward compatible, with the old setters marked as deprecated.

The idea with this is to treat capabilities differently based on how they need to go into the capabilities. 
`SauceOptions` --> top level / w3c compliant parameters
`SauceLabsOptions` --> parameters that go inside `sauce:options`

Then we'll add:
VisualOptions --> parameters that go inside `sauce:visual`
Then maybe we'll add (ideally we have users do what they do with Browsers & Selenium pass them in as constructors, but perhaps the ones we've been promoting on our site need to be able to be added directly)
AppiumOptions --> parameters prepended with `appium:`

Deprecated methods are being tested in `SauceOptionsDeprecatedTest.java`
`SauceOptionsTest.java` tests have all been updated to the new approach.

### DIFF
--- a/java/src/main/java/com/saucelabs/saucebindings/InvalidSauceOptionsArgumentException.java
+++ b/java/src/main/java/com/saucelabs/saucebindings/InvalidSauceOptionsArgumentException.java
@@ -1,7 +1,0 @@
-package com.saucelabs.saucebindings;
-
-class InvalidSauceOptionsArgumentException extends RuntimeException {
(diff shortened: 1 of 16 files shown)


nadvolod (approved):
Good work!

titusfortner:
closing as part of #246

PR #239, opened 2020-11-16
[java] implement factory / builder / config pattern

titusfortner:
(This is in Draft because it is based on #237)

This is the full implementation of #236, with all the browsers, so read the description there first


titusfortner:
Implemented in #246

PR #238, opened 2020-11-16
[java] implement visual options as a subclass

titusfortner:
(This is in DRAFT because it relies on #237)

This is a simple extension of what was done to the Sauce Labs Options class.

direct methods are w3c fields
methods accessed via `sauce()` are the ones using `sauce:options`
methods accessed via `visual()` are the ones using `sauce:visual`

```
 SauceOptions sauceOptions = new SauceOptions();
        sauceOptions.setBrowserVersion("77");
        sauceOptions.sauce().setBuild("Build Name");
        sauceOptions.visual().setProjectName("setsVisualValues");
        sauceOptions.visual().setViewportSize("1900x1200");
```

PR #240, opened 2020-11-16
Visual builder

titusfortner:
(DRAFT because it is a mix of #238 & #239 with the addition of visual builder functionality)

We can now do:
```
SauceOptions sauceOptions = SauceOptionsFactory.chrome()
                .setBrowserVersion("80")
                .visual()
                    .setProjectName("Project Name")
                    .build()
                .sauce()
                    .setBuild("My Build")
                    .build()
                .build();
```

VisualOptions is less complicated because everything is supported on all the platforms.
I'm not sure exactly how this looks when combined with mobile?

Also, tbh, I'm not certain what I'm doing with the default method on the interface, but it seems to work, so ¯\\_(ツ)_/¯


nadvolod (commented):
I think we unanimously decided this option is not the best

titusfortner:
Superceded by #256

PR #147, opened 2020-04-13
WIP for mobile support

titusfortner:
This is mostly working, but not yet ready for prime time;

Working code with specs:
* EmuSim Browser
* EmuSim App
* Test Object Browser

It's going to need more unit tests for the options and ideally a new release of Appium bindings.

Feedback welcome.

Also, note that this works around the issue of most of our mobile endpoints not yet supporting w3c.
So this "works" but does not fulfill the goal of the bindings to generate future-compliant code. That said, this is *why* bindings are a good idea in that we can toggle over to generate the "correct" syntax as soon as Sauce supports it.

PR #245, opened 2021-01-23
Java reorg

titusfortner:
This is based on #230 (which I updated with some extra javadocs and removed some unnecessary bits)

Note that the old SauceOptions class extends the new SauceOptions class in the options package and is marked as deprecated. The fix is as simple as changing the import at the top, but now it is backward compatible.

I also added tests to make sure that the deprecated stuff still works.

Next steps:
1. Merge #237 code
2. Merge #243 code
3. Add Other browsers
4. Release 1.1
5. Add Visual Support
6. Release 1.2
6. Add Mobile Support
7. Release 1.3
7. Remove deprecations
8. Release 2.0

Next step is to merge #237 code into this, then #243, then #238, then #147


titusfortner:
This is part of #246

PR #246, opened 2021-01-23
Java version 1.1 Proposal

titusfortner:
This is what I'd like to release as v1.1 
The "new feature" for users is the builder pattern which ensures that the users will only see options that are valid for the configuration they choose to use.

This is based on #230 #245 & #237 and subsumes #243
Each of the preceding commits has its own PR if you want to look at them individually.

I think there are some rough spots in here still, so any recommendations would be greatly appreciated.

Future Plans (for all bindings):
v1.2 --> Visual support
v1.3 --> Mobile support
v2.0 --> Remove Deprecations


titusfortner:
note - I might want to pull back in the enum I made for ScreenResolution in #243

nadvolod (approved):

PR #255, opened 2021-05-27
[java] add java-docs

titusfortner:
This should be everything for the 1.1 release


titusfortner:
too slow I'm merging it in preparation for a release....

PR #182, opened 2020-09-25
Visual testing API

nadvolod:
This is PR designed to iron out the visual API. Once we agree on it we can work on making it happen.
Here is a real world example of a current usage of Screener for visual testing https://github.com/saucelabs-training/demo-java/blob/master/java8/visual-e2e/visuale2e.junit4.examples/src/test/java/com/saucedemo/VisualCrossPlatformTests.java

PR #256, opened 2021-05-27
[Java] Add support for visual tests

titusfortner:
Update: PR is now comparing against Main branch and is ready for final review.

Anyone have any issues with using `setVisualOptions()` instead of putting these methods directly in the `Config` classes like we're doing with `sauce:options` values? There are a few other ways this could be implemented, but the code in this PR is closest to what we're doing in `SauceOptions` class for easy maintenance. Mobile shouldn't have this issue since it won't be using the browser static methods.

~Next step is adding Session functionality.~

The goal here is to make it as simple as possible to do the most expected things, while still allowing users flexibility to override things as needed.

The main limitation is that `setName()` must be used with options, because I couldn't figure out a way to create a reasonable default. if we ever create JUnit/TestNG specific jars we might be able to do something easier.

* So minimum to get a visual test is either:

1. Use `setVisual()`
```java
SauceOptions sauceOptions = SauceOptions.chrome().setName("Test Name").setVisual().build();
SauceSession sauceSession = new SauceSession(sauceOptions);
RemoteWebDriver driver = sauceSession.start();
```

2. Use `SauceSession#setDataCenter()`
```java
SauceOptions sauceOptions = SauceOptions.chrome().setName("Test Name").build();
SauceSession sauceSession = new SauceSession(sauceOptions);
sauceSession.setDataCenter(DataCenter.VISUAL);
RemoteWebDriver driver = sauceSession.start();
```

* To specify additional Visual Options, use `setVisualOptions()`:
```java
VisualOptions visualOptions = new VisualOptions();
visualOptions.setViewportSize("1024x768");

SauceOptions sauceOptions = SauceOptions.chrome()
    .setName("Test Name")
    .setVisualOptions(visualOptions)
    .build();
```

* `projectName` defaults to `build` value, but can be overridden in visual options with `visualOptions.setProjectName("Project Name")`
* `@visual.init` defaults to `name` value, which is required, but can also be changed as necessary with `session.visual().init("Test Name")`
* If `branch` is not set in `VisualOptions`, we check for actual git branch; if git is not detected, it defaults to `"default"` (not `"(default)"`) to ensure link in stdout works.
* Snapshots are taken by `session.visual().takeSnapshot("Snapshot Name")`
* Results are printed to stdout at the end of the test, but may also be obtained with: `Map results = session.visual().getResults()`

### DIFF
--- a/java/src/main/java/com/saucelabs/saucebindings/SauceSession.java
+++ b/java/src/main/java/com/saucelabs/saucebindings/SauceSession.java
@@ -3,6 +3,7 @@ package com.saucelabs.saucebindings;
 import com.saucelabs.saucebindings.options.BaseConfigurations;
 import com.saucelabs.saucebindings.options.InvalidSauceOptionsArgumentException;
 import com.saucelabs.saucebindings.options.SauceOptions;
+import com.saucelabs.saucebindings.options.VisualOptions;
 import lombok.Getter;
 import lombok.Setter;
 import org.openqa.selenium.InvalidArgumentException;
@@ -11,7 +12,6 @@ import org.openqa.selenium.remote.RemoteWebDriver;
 
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.time.Duration;
 
 public class SauceSession {
     @Getter @Setter private DataCenter dataCenter = DataCenter.US_WEST;
@@ -39,23 +39,27 @@ public class SauceSession {
     public SauceSession(SauceOptions options) {
         sauceOptions = options;
         if (sauceOptions.visual() != null) {
-            isVisualSession = true;
-            visualSession = new VisualSession(sauceOptions.visual());
+            dataCenter = DataCenter.VISUAL;
+            setVisual();
         }
     }
 
     public RemoteWebDriver start() {
-        if (isVisualSession) {
-            if (sauceOptions.sauce().getName() == null) {
-                String msg = "Visual Tests Require setting a name in options: SauceOptions#setName(Name)";
-                throw new InvalidSauceOptionsArgumentException(msg);
-            } else {
-                driver = createRemoteWebDriver(visualSession.getUrl(), sauceOptions.toCapabilities());
-                visualSession.start(driver, sauceOptions.sauce().getName());
-            }
-        } else {
-            driver = createRemoteWebDriver(getSauceUrl(), sauceOptions.toCapabilities());
+        if (dataCenter.equals(DataCenter.VISUAL) && sauceOptions.visual() == null) {
+            sauceOptions.setVisualOptions(new VisualOptions());
+            setVisual();
         }
+
+        if (isVisualSession && sauceOptions.sauce().getName() == null) {
+            String msg = "Visual Tests Require setting a name in options: SauceOptions#setName(Name)";
+            throw new InvalidSauceOptionsArgumentException(msg);
+        }
+        driver = createRemoteWebDriver(getSauceUrl(), sauceOptions.toCapabilities());
+
+        if (isVisualSession) {
+            visualSession.start(driver, sauceOptions.sauce().getName());
+        }
+
         return driver;
 	}
 
@@ -113,11 +117,8 @@ public class SauceSession {
         driver.quit();
     }
 
-    public void sleep(Duration time) {
-        try {
-            Thread.sleep(time.toMillis());
-        } catch (InterruptedException e) {
-            e.printStackTrace();
-        }
+    private void setVisual() {
+        isVisualSession = true;
+        visualSession = new VisualSession(sauceOptions.visual());
     }
 }
--- a/java/src/main/java/com/saucelabs/saucebindings/SauceSession.java
+++ b/java/src/main/java/com/saucelabs/saucebindings/SauceSession.java
@@ -1,6 +1,7 @@
 package com.saucelabs.saucebindings;
 
 import com.saucelabs.saucebindings.options.BaseConfigurations;
+import com.saucelabs.saucebindings.options.InvalidSauceOptionsArgumentException;
 import com.saucelabs.saucebindings.options.SauceOptions;
 import lombok.Getter;
 import lombok.Setter;
@@ -10,11 +11,14 @@ import org.openqa.selenium.remote.RemoteWebDriver;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.time.Duration;
 
 public class SauceSession {
     @Getter @Setter private DataCenter dataCenter = DataCenter.US_WEST;
     @Getter private final SauceOptions sauceOptions;
     @Setter private URL sauceUrl;
+    private VisualSession visualSession;
+    @Getter private Boolean isVisualSession = false;
 
     @Getter private RemoteWebDriver driver;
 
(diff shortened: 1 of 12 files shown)


c3ccl3ston:
@titusfortner, yeah, I likely still needed to be added to the `saucelabs` repos team or structure. @nadvolod had to do the same for me for the `saucelabs-training` repos

titusfortner:
Ok, I've updated the code, this should be good for merging. I have one more feature I'd like to add before we release 1.2, and we need to update the website documentation with info on how to use this, but this is the main chunk.

titusfortner:
I was looking at #182 
Another option is instead of having `VisualSession` instance be stored in a property in `SauceSession`, we could have it subclass `SauceSession`...

```java
visualSession = new VisualSession(sauceOptions);
RemoteWebDriver driver = visualSession.start();
visualSession.takeSnapshot("Name");
visualSession.getResults();
visualSession.stop("passed");
```

We could also overload it so that if you don't want to use an instance of `SauceOptions`, we could initialize it with Project Name:
```java
visualSession = new VisualSession("Project Name");
```
@nadvolod I'm guessing you might prefer this?

titusfortner:
Also of note, in doing this I split the `getBuildName()` into `defaultBuildName()` and `defaultBuildNumber()`; I also set `defaultBuildNumber()` to store the time in a static variable, so that when running locally it actually all ends up with the same build name instead of named with whatever timestamp each test gets run. Not sure why I didn't think of this long before.

java/src/test/java/com/saucelabs/saucebindings/integration/VisualTest.java
```diff
@@ -0,0 +1,116 @@
...
+        webDriver = session.start();
+        session.visual().takeSnapshot("Name of Snapshot");
+        Map results = session.visual().getResults();
+        Map snapshot = (Map) ((List) results.get("states")).get(0);
+        assertEquals("Name of Snapshot", snapshot.get("name"));
+    }
+
+    @Test
+    public void providesTestResults() {
+        SauceOptions options = SauceOptions.chrome().setName("Provides Test Results").setVisual().build();
```
nadvolod, line 80: @titusfortner if testName is mandatory, can't we force that as a parameter in `setVisual(string testName)` instead of throwing a run time exception if you forget `setName("Provides Test Results")`?

java/src/test/java/com/saucelabs/saucebindings/integration/VisualTest.java
```diff
@@ -0,0 +1,116 @@
...
+        session = new SauceSession(options);
+        webDriver = session.start();
+        session.visual().takeSnapshot("Name-of-Snapshot");
```
nadvolod, line 83: Interesting usage. Have we considered `session` just being a type of `VisualSession` that allows access to the appropriate visual methods? So we can do
```
VisualSession session = new SauceSession();
session.takeSnapshot("snapshotName");
```
or maybe
appropriate visual methods? So we can do
```
VisualSession session = new VisualSession();
session.takeSnapshot("snapshotName");
```

java/src/test/java/com/saucelabs/saucebindings/integration/VisualTest.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Test
+    public void acceptsNameForInit() {
+        SauceOptions options = SauceOptions.chrome().setName("Visual does not use this name").setVisual().build();
+        session = new SauceSession(options);
+        webDriver = session.start();
+        session.visual().init("Visual uses this name instead");
+        session.visual().takeSnapshot("Snapshot necessary to for Visual test to pass");
+        Map results = session.visual().getResults();
```
nadvolod, line 62: Can we provide a convenience method to get the status of the visual run? `session.visual().getTestStatus()`

java/src/main/java/com/saucelabs/saucebindings/examples/VisualTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.Map;
+
+public class VisualTest {
+
+    @Test
+    public void startSession() {
+        // 1. Create Visual Options instance
+        VisualOptions visualOptions = new VisualOptions();
+        visualOptions.setProjectName("My Project")
```
nadvolod, line 18: I think it's not very user friendly to force a user to `setProjectname()` with a run time exception instead of just forcing it through Intellisense. Have we considered:
`VisualOptions visualOptions = new VisualOptions("My Project");`

java/src/main/java/com/saucelabs/saucebindings/SystemManager.java
```diff
@@ -34,4 +38,20 @@ public static String get(String key) {
             return null;
         }
     }
+
+    public static String getCurrentGitBranch() {
```
nadvolod, line 42: What's the use case of this?

java/src/main/java/com/saucelabs/saucebindings/SauceSession.java
```diff
@@ -34,54 +38,87 @@ public SauceSession(BaseConfigurations configs) {
...
+                this.sauceOptions.sauce().getName());
+        String sauceTestLink = String.format("Test Job Link: https://app.saucelabs.com/tests/%s",
+                this.driver.getSessionId());
+        System.out.print(sauceReporter + "\n" + sauceTestLink + "\n");
     }
 
     private void stop() {
-        if(driver !=null) {
-            driver.quit();
+        if (isVisualSession) {
```
nadvolod, line 113: Shouldn't `stop()` also call `executeScript("/*@visual.end*/")`?

java/src/main/java/com/saucelabs/saucebindings/SauceSession.java
```diff
@@ -34,54 +38,87 @@ public SauceSession(BaseConfigurations configs) {
...
+        }
     }
 
     public RemoteWebDriver start() {
+        if (dataCenter.equals(DataCenter.VISUAL) && sauceOptions.visual() == null) {
+            sauceOptions.setVisualOptions(new VisualOptions());
+            setVisual();
+        }
+
+        if (isVisualSession && sauceOptions.sauce().getName() == null) {
```
nadvolod, line 53: I wonder if this can be simplified by forcing the user to set test name in the constructor. I talk about it below...

nadvolod (commented):
Added some comments. I'm also going to pull this code down to try it out

java/src/test/java/com/saucelabs/saucebindings/integration/VisualTest.java
titusfortner: Yes, I had an implementation that supported this in a couple different ways... but then I thought that ~90% of our users already set a name in their tests, and I'd rather encourage the other 10% to do it as well, rather than add extra code.

java/src/test/java/com/saucelabs/saucebindings/integration/VisualTest.java
titusfortner: yeah, you're going to like #259

java/src/test/java/com/saucelabs/saucebindings/integration/VisualTest.java
titusfortner: You mean convenience methods for parsing results?
Oh, we should make a `VisualResults` object and populate it in `getResults()` and then let the user query that rather than needing to parse the `Map`

java/src/main/java/com/saucelabs/saucebindings/SystemManager.java
titusfortner: To populate the `branch` value. If a value is not set, screener defaults to `"(default)"` (with the parentheses) which breaks some url renderings. This matches the behavior of screener with CLI, etc, and users can always override this value.

java/src/main/java/com/saucelabs/saucebindings/SauceSession.java
titusfortner: That's what the following line of code does when it prints (`visual().getResults()`)

titusfortner:
Ok, I think we've agreed that #259 is better, so let's move any conversations to that one.

PR #258, opened 2021-05-30
add accessibility functionality

titusfortner:
So this is kind of interesting.~

Sauce is working on getting an integration with Deque, so I thought it would be nice to add it to Sauce Bindings.
The examples for Sauce integration are all webdriver.io, so having the ability to point to easy usage in Java/Ruby/Python would be especially useful.

~I'm waiting to get access to the accessibility tab on Sauce so I can see what it is doing.
This runs the code, but it is too "chatty" ([this test](https://app.saucelabs.com/tests/accca70ee6ac4550996a3f0433d66f7e) is *with the iframes turned off). I might see if I can pull out just the commands we're interested in. It's doing a lot of things with frames that probably aren't necessary. I'm curious what a run with wdio looks like.~

~The configuration is also a little weird. I'm thinking we could do something like #256 with `AccessibilityOptions` class and a `setAccessibility(accessibilityOptions)` toggle in the options. I don't think we need anything else in Session other than these 2 methods.~

Update:
* Accessibility tab is turned on for Sauce users.
* All necessary configuration options are passed in to the `accessibilityResults()` method in the Session instance
* The code for Ruby & Python uses [sa11y](https://github.com/saucelabs/sa11y), which I wrote to be an efficient minimalist implementation with axe-core. It does not support any of the advanced features Axe provides.
* The code for Java is just a wrapper for the [Deque axe-core-maven-html library](https://github.com/dequelabs/axe-core-maven-html/); this makes it much more configurable than the sa11y implementations, but it is also less efficient than it could be.
* The Python code has been tested with a local copy of sa11y, but it needs to be released on pypi still


website/docs/ACCESSIBILITY.md
```diff
@@ -0,0 +1,57 @@
...
+---
+
+As of version 1.2, Sauce Bindings supports the new Sauce Labs 
+[Deque axe™ Integration](https://docs.saucelabs.com/basics/integrations/deque/index.html)
+
+* The Session classes now have a method to obtain accessibility results.
+* This does two basic things:
+    * Populates the new Accessibility tab in the Sauce Labs UI for a given job 
+      with the accessibility violations found during the test run
+    * Return the specific violations found in the code. The format of these results vary by language, and it is left to the user
```
Tacktician:
```suggestion
    * Returns the specific violations found in the code. The format of these results vary by language, and it is left to the user
```

website/docs/ACCESSIBILITY.md
```diff
@@ -0,0 +1,57 @@
...
+[Deque axe™ Integration](https://docs.saucelabs.com/basics/integrations/deque/index.html)
+
+* The Session classes now have a method to obtain accessibility results.
+* This does two basic things:
+    * Populates the new Accessibility tab in the Sauce Labs UI for a given job 
+      with the accessibility violations found during the test run
+    * Return the specific violations found in the code. The format of these results vary by language, and it is left to the user 
+      to determine how best to make use of them in their code if so desired.
+* The Java code is implemented using Deque's [Axe Core Maven HTML Jar](https://github.com/dequelabs/axe-core-maven-html)
+    * This is slightly less efficient than the other implementations, but provides access to all the advanced features
```
Tacktician:
```suggestion
    * This is slightly less efficient than the other implementations, but provides access to all the advanced features.
```

java/main/src/main/java/com/saucelabs/saucebindings/examples/AccessibilityTest.java
```diff
@@ -0,0 +1,35 @@
...
+        SauceSession session = new SauceSession();
+
+        // 2. Start Session to get the Driver
+        RemoteWebDriver driver = session.start();
+
+        // 3. Use the driver in your tests just like normal
+        driver.get("https://www.saucedemo.com/");
+
+        // 4a. Get accessibility default results with frame support
+        session.accessibilityResults();
```
nadvolod: This usage confused me some. If we are getting some results, would it make sense for the method to be called `getAccessibilityResults()`? And if we're getting the results, does it make sense to store those into a variable and/or do something with those results? Sorry if these are dumb questions. This is literally the first time I'm looking at this code and this feature in Sauce and these are the questions that are coming up.

java/main/src/main/java/com/saucelabs/saucebindings/examples/AccessibilityTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+        // 4b. Get accessibility default results without frame support
+        session.accessibilityResults(false);
```
nadvolod: This one is just as confusing as above and adding a boolean as a param just adds to the confusion about what I'm actually making false?

java/main/src/test/java/com/saucelabs/saucebindings/integration/AccessibilityTest.java
```diff
@@ -0,0 +1,68 @@
...
+            session.stop(true);
+        }
+    }
+
+    @Test
+    public void analyzesPageWithNestedIFrames() {
+        webDriver = session.start();
+        webDriver.get("http://watir.com/examples/nested_iframes.html");
+
+        Results results = session.accessibilityResults();
```
nadvolod: This is making more sense now. Still I think a better method name is `getAccessibilityResults()`

java/main/src/test/java/com/saucelabs/saucebindings/integration/AccessibilityTest.java
```diff
@@ -0,0 +1,68 @@
...
+        List<Rule> violations = results.getViolations();
+        int problems = violations.stream().map(e -> e.getNodes().size()).reduce(0, Integer::sum);
+        assertEquals(16, problems);
```
nadvolod, line 33: Again, new to automated accessibility testing. If this is the typical usage and users are validating that no violations exist, does it make sense to provide a simpler method for users:
`assertEquals(0, results.getViolationsCount())` which will do `violations.stream().map(e -> e.getNodes().size()).reduce(0, Integer::sum)` inside of `getViolationsCount()`

java/main/src/main/java/com/saucelabs/saucebindings/examples/AccessibilityTest.java
titusfortner: Oh, no, in Java it should totally be `getAccessibilityResults`.
So, the code returns a Java object that you can parse in your code... 
Or, you can just look at the results in the Accessibility tab in the Sauce UI, so you don't *have to do anything with the results. You can also call the method multiple times (presumably on multiple pages), so it doesn't make sense to store anything in the session class.

java/main/src/main/java/com/saucelabs/saucebindings/examples/AccessibilityTest.java
titusfortner: The parameter is for turning off extra frame commands if you don't have frames on the page. 
In Ruby & Python you get keywords to make it more obvious that's what it is doing. In Java you have to rely on the IDE.

java/main/src/test/java/com/saucelabs/saucebindings/integration/AccessibilityTest.java
titusfortner: The functionality is mostly just to populate the tab in the UI. There are better ways to test this SB code is working as intended, this was the "easy" way for me to make sure it was capturing exactly what I expected w/ and w/o frames. If users are interested in a more convenient Results class, we can add one. I just didn't want to add it from the get go without knowing if people were going to be using it or if they cared.

PR #262, opened 2021-06-07
Test runners

titusfortner:
I'm merging #260 & #261 into this; we'll make sure it's all good then release the packages.

Note: this is a PR against `move_java` branch; need to update it to main before release.


titusfortner:
Merged this

PR #266, opened 2021-07-10
implement class methods for SauceOptions

titusfortner:
This matches the Java 1.1.0 code to limit options to only what is valid for the configuration

### DIFF
--- a/python/saucebindings/options.py
+++ b/python/saucebindings/options.py
@@ -1,8 +1,8 @@
-from selenium.webdriver.chrome.options import Options as ChromeOptions
-from selenium.webdriver.firefox.options import Options as FirefoxOptions
-from selenium.webdriver.edge.options import Options as EdgeOptions
-from selenium.webdriver.ie.options import Options as IEOptions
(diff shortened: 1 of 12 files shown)


python/saucebindings/options.py
```diff
@@ -130,8 +158,15 @@ def __getattr__(self, key):
         except KeyError:
             return None
 
+    def validateOptions(self, kwargs):
```
joshgrantsauce, line 161: Looks like method needs to be moved up to be defined before `__init__`

python/saucebindings/session.py
```diff
@@ -15,7 +15,7 @@
 class SauceSession():
 
     def __init__(self, options=None, data_center='us-west', resolve_ip=False):
-        self.options = options if options else SauceOptions()
+        self.options = options if options else SauceOptions.chrome()
```
joshgrantsauce, line 18: This is actually a bit better for being explicit, so good job here

joshgrantsauce (approved):

Issue #267, opened 2021-07-21
Capture Performance vs Test Name


titusfortner:
Right now Java errors if capture performance is set without setting test name, but with the test runners this is set automatically *after* the session is created, requiring users to set it. We should remove the check in the setter and put it in the capability conversion method.

nadvolod:
@titusfortner we should fix this bug, our current Performance example is weird as it's really tough to set a good test name per test as it's implemented in. 
```java
   @Override
    public SauceOptions createSauceOptions() {
        return SauceOptions.chrome()
                .setExtendedDebugging()
                .setName("PerformanceTest") // bug - https://github.com/saucelabs/sauce_bindings/issues/267
                .setCapturePerformance()
                .build();
    }
```
I didn't take a long look, not sure if you know of a work around?

PR #269, opened 2021-07-21
[java] validate capturePerformance requirements in build() 

titusfortner:
This addresses #267 
This should be a good pattern going forward of addressing incompatibilities in the overridden build() method.


nadvolod (commented):
@titusfortner love small PRs! The code makes sense.
1. Does it make sense to add a test for this case?
2. I see that the Checks are failing for Java 8 and 9
3. I see that there are a few Codacy complaints. Can you see those and are they valid in your opinion?

PR #270, opened 2021-07-22
[java] implement performance support in session

### LINKED ISSUE #247: Add Sauce performance enablement and disablement
New feature released from Sauce https://wiki.saucelabs.com/display/DOCSDEV/Measure+Page+Load+Performance+Using+Test+Automation
It's relatively easy to implement and allows a user to pause performance until a certain page.

```java
browser.execute('sauce:performanceDisable')
browser.url('https://www.json.org/json-en.html')
browser.execute('sauce:performanceEnable')
browser.url('https://example.com')
```

titusfortner:
Three options here.

The one I did for Accessibility was like:

```java
PerformanceResults results = session.getPerformanceResults();
```

The one that is probably "most Java" is:
```java
PerformanceResults results = new Performance(session).getResults();
```

The one I implemented here is:
```java
PerformanceResults results = session.performance().getResults();
```

Thoughts/Preferences?

### DIFF
--- a/java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
@@ -3,7 +3,9 @@ package com.saucelabs.saucebindings;
 import com.deque.html.axecore.results.Results;
 import com.deque.html.axecore.selenium.AxeBuilder;
 import com.saucelabs.saucebindings.options.BaseConfigurations;
+import com.saucelabs.saucebindings.options.InvalidSauceOptionsArgumentException;
 import com.saucelabs.saucebindings.options.SauceOptions;
+import com.saucelabs.saucebindings.performance.Performance;
 import lombok.Getter;
 import lombok.Setter;
 import org.openqa.selenium.InvalidArgumentException;
@@ -73,6 +75,13 @@ public class SauceSession {
 
     public Results getAccessibilityResults(AxeBuilder builder) {
         return builder.analyze(driver);
+    }
+
+    public Performance performance() {
+        if (!sauceOptions.sauce().getCapturePerformance()) {
+            throw new InvalidSauceOptionsArgumentException("Can not use performance methods without using `setCapturePerformance()`");
+        }
+        return new Performance(driver, sauceOptions.sauce().getName());
     }
 
     public void stop(Boolean passed) {
--- a/java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
@@ -81,7 +81,7 @@ public class SauceSession {
         if (!sauceOptions.sauce().getCapturePerformance()) {
             throw new InvalidSauceOptionsArgumentException("Can not use performance methods without using `setCapturePerformance()`");
         }
-        return new Performance(driver, sauceOptions.sauce().getName());
+        return new Performance(this);
     }
 
     public void stop(Boolean passed) {
--- a/java/main/src/main/java/com/saucelabs/saucebindings/performance/Performance.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/performance/Performance.java
@@ -58,4 +58,25 @@ public class Performance {
         driver.executeScript("sauce:performanceDisable");
     }
 
+    // https://docs.saucelabs.com/performance/transitions#defining-a-performance-budget
+    // Can pull this from yaml or json
+    public Map<String, Object> getBudgetFailures(Map<String, Map<String, Object>> budget) {
+        Map<String, Object> violations = new HashMap<>();
+
+        budget.forEach((url, values) -> {
+            driver.get(url);
+            PerformanceMetrics metrics = getMetrics();
+            values.forEach((metric, value) -> {
+                if ("score".equals(metric)) {
+                    Double score = (Double) metrics.getRawData().get(metric);
+                    if (score < (Double) value) {
+                        violations.put(metric, value);
+                    }
+                } else if (((Number) metrics.getRawData().get(metric)).longValue() > ((Number) value).longValue()) {
+                    violations.put(metric, value);
+                }
+            });
+        });
+        return violations;
+    }
 }
--- a/java/main/src/main/java/com/saucelabs/saucebindings/performance/Performance.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/performance/Performance.java
@@ -1,6 +1,6 @@
 package com.saucelabs.saucebindings.performance;
 
-import org.openqa.selenium.remote.RemoteWebDriver;
+import com.saucelabs.saucebindings.SauceSession;
 
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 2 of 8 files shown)


java/main/src/main/java/com/saucelabs/saucebindings/performance/Performance.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * The map used as input will typically be converted from serialized information stored in a yaml or JSON file
+     * @param budget Map of Maps with the URL as the keys of outer maps, the metric as the keys of the nested maps and the maximum value
+     *               of the metric as its value
+     * @return A Map of failures if any, including the failing metric and its value
+     * @see <a href="https://docs.saucelabs.com/performance/transitions#defining-a-performance-budget">Defining a Performance Budget</a>
+     */
+    public Map<String, Object> getBudgetFailures(Map<String, Map<String, Object>> budget) {
```
nadvolod, line 112: Not sure I understand what this method does when first reading it. The term `budget` really throws me off as I've never encountered it with front-end perf.
Is it meant to be public?

java/main/src/test/java/com/saucelabs/saucebindings/budget.yml
```diff
@@ -0,0 +1,6 @@
+"https://www.saucedemo.com":
```
nadvolod, line 1: @titusfortner can you please explain what this is and it's intended usage?

java/main/src/test/java/com/saucelabs/saucebindings/integration/PerformanceTest.java
```diff
@@ -0,0 +1,168 @@
...
+                .build();
+        session = new SauceSession(sauceOptions);
+        driver = session.start();
+    }
+
+    @Test
+    public void jankinessResults() {
+        driver.get("https://www.saucedemo.com");
+
+        JankinessResults jankinessResults = session.performance().getJankinessResults();
```
nadvolod, line 54: I like having this return a `JankinessResults` instead of having to parse through a Map. Definitely more user friendly

java/main/src/test/java/com/saucelabs/saucebindings/integration/PerformanceTest.java
```diff
@@ -0,0 +1,168 @@
...
+            Map<String, Object> firstInteractive = (Map<String, Object>) performanceResults.getDetails().get("firstInteractive");
+            Assert.assertTrue((Long) firstInteractive.get("actual") < 5000);
+        }
+    }
+
+    @Test
+    public void performanceDetails() {
+        driver.get("https://www.saucedemo.com");
+
+        PerformanceMetrics performanceMetrics = session.performance().getMetrics();
```
nadvolod, line 111: This confuses me. I don't understand the difference between a `PerformanceMetrics` and `PerformanceResults`

java/main/src/test/java/com/saucelabs/saucebindings/integration/PerformanceTest.java
```diff
@@ -0,0 +1,168 @@
...
+            session.stop(true);
+        }
+    }
+
+    @Rule
+    public TestName testName = new TestName();
+
+    @Before
+    public void setup() {
+        SauceOptions sauceOptions = SauceOptions.chrome()
```
nadvolod, line 42: Is `setCapturePerformance()` exposed on other browsers besides `chrome()`?

nadvolod (commented):
I like this approach
`PerformanceResults results = session.performance().getResults();`
I think that the API usage makes sense and I also like having the performance stored in a separate object `Performance`.

java/main/src/main/java/com/saucelabs/saucebindings/performance/Performance.java
titusfortner: Yeah, look at the link to our performance docs that I put in javadocs which discusses it. Christian implemented it for wdio, and showed JS code. Just trying to keep up with wdio. :)

java/main/src/test/java/com/saucelabs/saucebindings/integration/PerformanceTest.java
titusfortner: Nope. Chrome only.

java/main/src/test/java/com/saucelabs/saucebindings/integration/PerformanceTest.java
titusfortner: Maybe we rename PerformanceMetrics to PerformanceLog.

PerformanceResults uses Sauce Labs magic AI values. You can pass in specific metrics you care about and Sauce will pass/fail it based on whether we think it is in range.

If you want to assert on anything  based on a hard coded value instead of our magic, you need to first get the log.

I mean, maybe this should be:
```java
PerformanceLogs performanceLogs = session.logs().getPerformance();
```

java/main/src/test/java/com/saucelabs/saucebindings/budget.yml
titusfortner: So, for JS they use JSON; I wanted to implement something that uses a Map, so you can convert either yaml or json into the map it will work. I didn't want to use JSON because I don't want to deal with Jackson library, yet. :) More info: https://docs.saucelabs.com/performance/transitions#defining-a-performance-budget

nadvolod: @titusfortner I worry that this feature is unnecessarily making the code complicated and overengineered. From my experience, customers aren't advanced enough to handle such a use case (or even me 😂). Instead they are only interested in [3 use cases](https://github.com/saucelabs-training/demo-java/blob/master/selenium-junit4-examples/src/test/java/com/saucedemo/PerformanceExampleTests.java). Specifically they like `multiPagePerformanceExample()` and the ` simplePerformanceTest()`

java/main/src/test/java/com/saucelabs/saucebindings/integration/PerformanceTest.java
```diff
@@ -0,0 +1,168 @@
...
+            Assert.assertTrue(performanceResults.getReason().contains("Metrics"));
+            Assert.assertFalse(performanceResults.getDetails().isEmpty());
+        }
+    }
+
+    @Test
+    public void performanceMetricResults() {
+        driver.get("https://www.saucedemo.com");
+
+        PerformanceResults performanceResults = session.performance().getResults("firstInteractive");
```
nadvolod, line 77: @titusfortner should we move the parameter values into an enum? They've remained the same forever and it should improve usability

java/main/src/test/java/com/saucelabs/saucebindings/budget.yml
titusfortner: Yeah, We have docs for it with code examples, so I thought we should provide support for it. I'm not sure how much it would be used. I had some thoughts on expanding it later, but definitely not the thing that should be holding us up.

This is the kind of feature I wouldn't make an issue out of in the Sauce Bindings Documentation, but would have example code in the Sauce Docs with it for people who are interested in playing with it. As such, the confusion factor of the average user shouldn't really matter. It's the same kind of thing with the `mergeCapabilities()` in SauceOptions.

java/main/src/test/java/com/saucelabs/saucebindings/integration/PerformanceTest.java
nadvolod: Makes sense. The API call would be what you said or `session.performance().getLogs()`

titusfortner: The other part of this is that it isn't all the logs, just the logs for the last page transition. `session.logs()` should be getting all the logs for something, so maybe `session.performance().getLastLog()`?

java/main/src/test/java/com/saucelabs/saucebindings/integration/PerformanceTest.java
titusfortner: Yeah, so long as we have `getRawResults()` I don't think there's a problem doing enums for this.

java/main/src/test/java/com/saucelabs/saucebindings/integration/PerformanceTest.java
nadvolod: @titusfortner so if it's just for the last page transition then maybe?
`session.performance().getLastPageTransitionLog()`

java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
```diff
@@ -75,6 +77,13 @@ public Results getAccessibilityResults(AxeBuilder builder) {
         return builder.analyze(driver);
     }
 
+    public Performance performance() {
+        if (!sauceOptions.sauce().getCapturePerformance()) {
```
titusfortner, line 81: what does this do for non-Chrome which doesn't have this method?

java/main/src/main/java/com/saucelabs/saucebindings/performance/Performance.java
```diff
@@ -0,0 +1,131 @@
...
+        metrics.add(metric);
+        return getResults(metrics);
+    }
+
+    /**
+     * This measures the performance output of a specific list of metrics against a baseline of previously accepted performance values
+     *
+     * @param metrics A List of values for which performance metrics to evaluate
+     * @return PerformanceResults is a wrapper of the HashMap results to make it easier to work with them
+     * @see <a href="https://docs.saucelabs.com/performance/transitions#implementing-the-performance-command-assertion">Implementing the Performance Command Assertion</a>
```
titusfortner, line 55: these links are wrong for anchors, they need to be `/#` without `index.html`

java/main/src/main/java/com/saucelabs/saucebindings/performance/Performance.java
```diff
@@ -0,0 +1,131 @@
...
+     * @see SauceSession#performance()
+     */
+    public Performance(SauceSession session) {
+        this.session = session;
+    }
+
+    /**
+     * This measures the performance output against a baseline of previously accepted performance values
+     *
+     * @return PerformanceResults is a wrapper of the HashMap results to make it easier to work with them
```
titusfortner, line 26: specify this is `passed`, `reason`, and `details`

java/main/src/main/java/com/saucelabs/saucebindings/performance/Performance.java
```diff
@@ -0,0 +1,131 @@
...
+     * @see <a href="https://docs.saucelabs.com/performance/transitions#implementing-the-performance-command-assertion">Implementing the Performance Command Assertion</a>
+     */
+    public PerformanceResults getResults() {
```
titusfortner, line 29: probably needs to be `getAllResults()` to better differentiate from other method

PR #275, opened 2022-02-16
Java fixes

titusfortner:
These are some generic changes I tweaked while working on other things.

Should be relatively straightforward what's happening here

### DIFF
--- a/java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
@@ -1,0 +1,73 @@
+package com.saucelabs.saucebindings;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+
+/**
+ * Determines the information a test needs based on what CI Tool is being used.
+ * It does not need to be initialized.
+ */
+public abstract class CITools {
+    private static String buildName = "Default Build Name";
+    private static String buildNumber = String.valueOf(System.currentTimeMillis());
+    private static Boolean buildInfoStored = false;
+
+    /**
+     * Map of CI Tools and what Environment variable, if present, indicates its usage.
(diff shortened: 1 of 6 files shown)


java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
```diff
@@ -0,0 +1,58 @@
...
+            "GitLab", "CI",
+            "TeamCity", "TEAMCITY_PROJECT_NAME");
+
+    public static Map<String, ImmutableList<String>> buildValues = ImmutableMap.of(
+            "Jenkins", ImmutableList.of("BUILD_NAME", "BUILD_NUMBER"),
+            "Bamboo", ImmutableList.of("bamboo_shortJobName", "bamboo_buildNumber"),
+            "Travis", ImmutableList.of("TRAVIS_JOB_NAME", "TRAVIS_JOB_NUMBER"),
+            "Circle", ImmutableList.of("CIRCLE_JOB", "CIRCLE_BUILD_NUM"),
+            "GitLab", ImmutableList.of("CI_JOB_NAME", "CI_JOB_ID"),
+            "TeamCity", ImmutableList.of("TEAMCITY_PROJECT_NAME", "BUILD_NUMBER"));
```
nadvolod: @titusfortner should we add Github Actions since it's so popular now?

nadvolod (commented):
@titusfortner love the new updates. The code reads so much cleaner. Added one comment. Also, what do you think of those Codacy issues? Many can likely be ignored, but some seem valid to me. Let me know and I can work with you to fix this.

java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
titusfortner: oddly, we added it to Ruby & Python, but not Java? Maybe cut and paste removed somewhere? Either way, added it.

PR #276, opened 2022-02-16
Java visual

titusfortner:
This is a PR against #275, and it'll stay in Draft until that one is merged.
It subsumes #259 

I finally really like this API!

Easiest to see what this does from the examples:

Using default:
https://github.com/saucelabs/sauce_bindings/blob/11d774517a0b2c293c1c2c28939aa36375d1eeaa/java/main/src/test/java/com/saucelabs/saucebindings/examples/VisualTest.java#L12-L26

Updating Options:
https://github.com/saucelabs/sauce_bindings/blob/11d774517a0b2c293c1c2c28939aa36375d1eeaa/java/main/src/test/java/com/saucelabs/saucebindings/examples/VisualOptionsTest.java#L14-L39

Cool things this does:
* Automatically populates the default branch name with the actual branch name if it can determine it
* Automatically populates the Sauce Labs job with pass/fail based on whether Visual tests pass/fail
* Useful error when attempting to send a visual command after getting results (i.e., calling "end")

I still need to add javadocs because that's important and stuff

### DIFF
--- a/java/main/src/main/java/com/saucelabs/saucebindings/GitManager.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/GitManager.java
@@ -1,0 +1,24 @@
+package com.saucelabs.saucebindings;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+public class GitManager {
+    private static String currentBranch = "_default_";
+
+    public static String getCurrentBranch() {
+        try {
+            Process process = Runtime.getRuntime().exec("git rev-parse --abbrev-ref HEAD");
+            process.waitFor();
+            BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
+            String line = reader.readLine();
+            if (!line.contains("fatal:")) {
+                currentBranch = line;
+            }
+        } catch (IOException | InterruptedException e) {
+            // Ignore exception and use default
+        }
+        return currentBranch;
+    }
+}
--- a/java/main/src/main/java/com/saucelabs/saucebindings/VisualResults.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/VisualResults.java
@@ -1,0 +1,53 @@
+package com.saucelabs.saucebindings;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.openqa.selenium.remote.RemoteWebDriver;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+@Setter @Getter
+public class VisualResults {
+    Map<String, Object> results;
+    private Map<String, Object> totals;
+
+    private final Boolean passed;
+    private final String message;
+    private final String status;
+    private final List<VisualSnapshot> snapshots = new ArrayList<>();
+    private final Long totalNew;
+    private final Long totalChanged;
+    private final Long totalRejected;
+    private final Long totalAccepted;
+    private final Long total;
+
+    public static VisualResults generate(RemoteWebDriver driver) {
+        return new VisualResults(driver);
+    }
+
+    private VisualResults(RemoteWebDriver driver) {
+        results = (Map<String, Object>) driver.executeScript("/*@visual.end*/");
+
+        this.passed = (Boolean) results.get("passed");
+        this.message = (String) results.get("message");
+        this.status = (String) results.get("status"); // success, failure, error, timeout, cancelled
+
+        for (Map state : (List<Map>) results.get("states")) {
+            snapshots.add(new VisualSnapshot(state));
+        }
+
+        this.totals = (Map) results.get("totals");
+        this.totalNew = (Long) totals.get("new");
+        this.totalChanged = (Long) totals.get("changed");
+        this.totalRejected = (Long) totals.get("rejected");
+        this.totalAccepted = (Long) totals.get("accepted");
+        this.total = (Long) totals.get("all");
+    }
+
+    public String getSummary() {
+        String disposition = getPassed() ? getStatus() : getMessage();
+        return "Totals: " + getTotals() + "; Disposition: " + disposition + "\n";
+    }
+}
--- a/java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
@@ -1,18 +1,25 @@
 package com.saucelabs.saucebindings;
 
 import com.saucelabs.saucebindings.options.SauceOptions;
+import com.saucelabs.saucebindings.options.VisualOptions;
 import org.openqa.selenium.InvalidArgumentException;
 import org.openqa.selenium.MutableCapabilities;
 import org.openqa.selenium.remote.RemoteWebDriver;
 
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.HashMap;
 import java.util.Map;
 
 public class VisualSession extends SauceSession {
+    private VisualOptions visualOptions;
+
     public VisualSession(String testName) {
-        super(SauceOptions.chrome().setName(testName).build());
+        this(new VisualOptions(SauceOptions.chrome().setName(testName).build()));
+    }
+
+    public VisualSession(VisualOptions options) {
+        super(options.getSauceOptions());
+        this.visualOptions = options;
     }
 
     @Override
@@ -25,15 +32,7 @@ public class VisualSession extends SauceSession {
     }
 
     public RemoteWebDriver start() {
-        MutableCapabilities capabilities = getSauceOptions().toCapabilities();
-
-        String sauceVisualApiKey = SystemManager.get("SCREENER_API_KEY");
-        Map<String, Object> visual = new HashMap<>();
-        visual.put("apiKey", sauceVisualApiKey);
-        visual.put("projectName", new CITools().getBuildName());
-        visual.put("branch", "_default_");
-
-        capabilities.setCapability("sauce:visual", visual);
+        MutableCapabilities capabilities = visualOptions.toCapabilities();
 
         this.driver = createRemoteWebDriver(getSauceUrl(), capabilities);
         driver.executeScript("/*@visual.init*/", getSauceOptions().sauce().getName());
--- a/java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
@@ -1,0 +1,56 @@
+package com.saucelabs.saucebindings;
+
+import com.saucelabs.saucebindings.options.SauceOptions;
+import org.openqa.selenium.InvalidArgumentException;
+import org.openqa.selenium.MutableCapabilities;
+import org.openqa.selenium.remote.RemoteWebDriver;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.HashMap;
+import java.util.Map;
+
+public class VisualSession extends SauceSession {
+    public VisualSession(String testName) {
+        super(SauceOptions.chrome().setName(testName).build());
+    }
+
+    @Override
+    public URL getSauceUrl() {
+        try {
+            return new URL("https://hub.screener.io/wd/hub");
+        } catch (MalformedURLException e) {
+            throw new InvalidArgumentException("Invalid URL", e);
+        }
+    }
+
+    public RemoteWebDriver start() {
+        MutableCapabilities capabilities = getSauceOptions().toCapabilities();
+
+        String sauceVisualApiKey = SystemManager.get("SCREENER_API_KEY");
+        Map<String, Object> visual = new HashMap<>();
+        visual.put("apiKey", sauceVisualApiKey);
+        visual.put("projectName", new CITools().getBuildName());
+        visual.put("branch", "_default_");
+
+        capabilities.setCapability("sauce:visual", visual);
+
+        this.driver = createRemoteWebDriver(getSauceUrl(), capabilities);
+        driver.executeScript("/*@visual.init*/", getSauceOptions().sauce().getName());
+
+        return driver;
+    }
+
+    public void takeSnapshot(String name) {
+        driver.executeScript("/*@visual.snapshot*/", name);
+    }
+
+    public Map<String, Object> getResults() {
+        return (Map<String, Object>) driver.executeScript("/*@visual.end*/");
+    }
+
+    public void stop() {
+        System.out.println("\n Visual Results: " + getResults());
+        super.quit();
+    }
+}
--- a/java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
@@ -3,15 +3,14 @@ package com.saucelabs.saucebindings;
 import com.saucelabs.saucebindings.options.SauceOptions;
 import com.saucelabs.saucebindings.options.VisualOptions;
 import org.openqa.selenium.InvalidArgumentException;
-import org.openqa.selenium.MutableCapabilities;
 import org.openqa.selenium.remote.RemoteWebDriver;
 
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.Map;
 
 public class VisualSession extends SauceSession {
-    private VisualOptions visualOptions;
+    private VisualResults visualResults;
+    private final VisualOptions visualOptions;
 
     public VisualSession(String testName) {
         this(new VisualOptions(SauceOptions.chrome().setName(testName).build()));
@@ -23,6 +22,13 @@ public class VisualSession extends SauceSession {
     }
 
     @Override
+    public RemoteWebDriver start() {
+        this.driver = createRemoteWebDriver(getSauceUrl(), visualOptions.toCapabilities());
+        newVisualTest(getSauceOptions().sauce().getName());
+        return driver;
+    }
+
+    @Override
     public URL getSauceUrl() {
         try {
             return new URL("https://hub.screener.io/wd/hub");
@@ -31,25 +37,49 @@ public class VisualSession extends SauceSession {
         }
     }
 
-    public RemoteWebDriver start() {
-        MutableCapabilities capabilities = visualOptions.toCapabilities();
-
-        this.driver = createRemoteWebDriver(getSauceUrl(), capabilities);
-        driver.executeScript("/*@visual.init*/", getSauceOptions().sauce().getName());
-
-        return driver;
+    public void newVisualTest(String testName) {
+        validateVisualStatus();
+        driver.executeScript("/*@visual.init*/", testName);
     }
 
     public void takeSnapshot(String name) {
+        validateVisualStatus();
         driver.executeScript("/*@visual.snapshot*/", name);
     }
 
-    public Map<String, Object> getResults() {
-        return (Map<String, Object>) driver.executeScript("/*@visual.end*/");
+    public VisualResults getVisualResults() {
+        if (visualResults == null) {
+            this.visualResults = VisualResults.generate(driver);
+        }
+        return visualResults;
+    }
+
+    @Override
+    public void stop(Boolean passed) {
+        String result = passed ? "passed" : "failed";
+        stop(result);
+    }
+
+    @Override
+    public void stop(String result) {
+        if (driver != null) {
+            System.out.println(getVisualResults().getSummary());
+            super.stop(result);
+        }
     }
 
     public void stop() {
-        System.out.println("\n Visual Results: " + getResults());
-        super.quit();
+        if (driver != null) {
+            VisualResults results = getVisualResults();
+            System.out.println(results.getSummary());
+            String result = results.getPassed() ? "passed" : "failed";
+            super.stop(result);
+        }
+    }
+
+    private void validateVisualStatus() {
+        if (this.visualResults != null) {
+            throw new SauceVisualException("Can not execute Visual Commands after calling getVisualResults()");
+        }
     }
 }
--- a/java/main/src/main/java/com/saucelabs/saucebindings/options/VisualOptions.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/options/VisualOptions.java
@@ -1,6 +1,7 @@
 package com.saucelabs.saucebindings.options;
 
 import com.saucelabs.saucebindings.CITools;
+import com.saucelabs.saucebindings.GitManager;
 import com.saucelabs.saucebindings.SauceSession;
 import com.saucelabs.saucebindings.SystemManager;
 import lombok.AccessLevel;
@@ -73,7 +74,7 @@ public class VisualOptions extends BaseOptions {
         }
 
         if (branch == null) {
-            branch = "_default_";
+            branch = GitManager.getCurrentBranch();
         }
 
         capabilityManager.addCapabilities();
--- a/java/main/src/main/java/com/saucelabs/saucebindings/options/VisualOptions.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/options/VisualOptions.java
@@ -1,0 +1,86 @@
+package com.saucelabs.saucebindings.options;
+
+import com.saucelabs.saucebindings.CITools;
+import com.saucelabs.saucebindings.SauceSession;
+import com.saucelabs.saucebindings.SystemManager;
+import lombok.AccessLevel;
+import lombok.Getter;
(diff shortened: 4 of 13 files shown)


java/main/src/main/java/com/saucelabs/saucebindings/GitManager.java
```diff
@@ -0,0 +1,24 @@
+package com.saucelabs.saucebindings;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+public class GitManager {
+    private static String currentBranch = "_default_";
```
mykola-mokhnach, line 8: does such branch (_default_) really exist?

java/main/src/main/java/com/saucelabs/saucebindings/GitManager.java
```diff
@@ -0,0 +1,24 @@
...
+            process.waitFor();
+            BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
+            String line = reader.readLine();
+            if (!line.contains("fatal:")) {
+                currentBranch = line;
+            }
+        } catch (IOException | InterruptedException e) {
+            // Ignore exception and use default
+        }
+        return currentBranch;
```
mykola-mokhnach, line 22: is it ok that a previously remembered value is returned if an exception happens?

java/main/src/main/java/com/saucelabs/saucebindings/GitManager.java
```diff
@@ -0,0 +1,24 @@
...
+    private static String currentBranch = "_default_";
+
+    public static String getCurrentBranch() {
+        try {
+            Process process = Runtime.getRuntime().exec("git rev-parse --abbrev-ref HEAD");
+            process.waitFor();
+            BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
+            String line = reader.readLine();
+            if (!line.contains("fatal:")) {
+                currentBranch = line;
```
mykola-mokhnach, line 17: should the line be trimmed?

java/main/src/main/java/com/saucelabs/saucebindings/VisualResults.java
```diff
@@ -0,0 +1,53 @@
...
+    public static VisualResults generate(RemoteWebDriver driver) {
+        return new VisualResults(driver);
+    }
+
+    private VisualResults(RemoteWebDriver driver) {
+        results = (Map<String, Object>) driver.executeScript("/*@visual.end*/");
+
+        this.passed = (Boolean) results.get("passed");
+        this.message = (String) results.get("message");
+        this.status = (String) results.get("status"); // success, failure, error, timeout, cancelled
```
mykola-mokhnach, line 35: does it make sense to have an enum instead?

java/main/src/main/java/com/saucelabs/saucebindings/VisualResults.java
```diff
@@ -0,0 +1,53 @@
...
+
+        for (Map state : (List<Map>) results.get("states")) {
```
mykola-mokhnach, line 37: this could also be rewritten in more functional style:

```
((List<Map>) results.get("states")).map(VisualSnapshot::new).forEach(snapshots::add)
```

java/main/src/main/java/com/saucelabs/saucebindings/VisualResults.java
```diff
@@ -0,0 +1,53 @@
...
+import lombok.Setter;
+import org.openqa.selenium.remote.RemoteWebDriver;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+@Setter @Getter
+public class VisualResults {
+    Map<String, Object> results;
```
mykola-mokhnach, line 13: should the field be private?

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
```diff
@@ -0,0 +1,85 @@
...
+    public RemoteWebDriver start() {
+        this.driver = createRemoteWebDriver(getSauceUrl(), visualOptions.toCapabilities());
+        newVisualTest(getSauceOptions().sauce().getName());
+        return driver;
+    }
+
+    @Override
+    public URL getSauceUrl() {
+        try {
+            return new URL("https://hub.screener.io/wd/hub");
```
mykola-mokhnach, line 34: should the URL be moved into a constant/config?

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public void stop(Boolean passed) {
+        String result = passed ? "passed" : "failed";
+        stop(result);
+    }
+
+    @Override
+    public void stop(String result) {
+        if (driver != null) {
+            System.out.println(getVisualResults().getSummary());
```
mykola-mokhnach, line 66: do we really want to use println? this function does not allow any external configuration/customization

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
```diff
@@ -0,0 +1,85 @@
...
+    public VisualResults getVisualResults() {
+        if (visualResults == null) {
+            this.visualResults = VisualResults.generate(driver);
+        }
+        return visualResults;
+    }
+
+    @Override
+    public void stop(Boolean passed) {
+        String result = passed ? "passed" : "failed";
```
mykola-mokhnach, line 59: magic strings

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public URL getSauceUrl() {
+        try {
+            return new URL("https://hub.screener.io/wd/hub");
+        } catch (MalformedURLException e) {
+            throw new InvalidArgumentException("Invalid URL", e);
+        }
+    }
+
+    public void newVisualTest(String testName) {
```
mykola-mokhnach, line 40: I like when methods return `this` instead of `void`, so calls could be chained

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
```diff
@@ -0,0 +1,85 @@
...
+            System.out.println(getVisualResults().getSummary());
+            super.stop(result);
+        }
+    }
+
+    public void stop() {
+        if (driver != null) {
+            VisualResults results = getVisualResults();
+            System.out.println(results.getSummary());
+            String result = results.getPassed() ? "passed" : "failed";
```
mykola-mokhnach, line 75: I can observe some code duplication here

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
mykola-mokhnach, line 40: should this method be public? can we start multiple tests in scope of a single session?

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public VisualSession(VisualOptions options) {
+        super(options.getSauceOptions());
+        this.visualOptions = options;
+    }
+
+    @Override
+    public RemoteWebDriver start() {
+        this.driver = createRemoteWebDriver(getSauceUrl(), visualOptions.toCapabilities());
```
mykola-mokhnach, line 26: I assume `this. ` is redundant

java/main/src/main/java/com/saucelabs/saucebindings/options/VisualOptions.java
```diff
@@ -0,0 +1,105 @@
...
+    private String ignore;
+    private Boolean failOnNewStates = null;
+    private Boolean alwaysAcceptBaseBranch = null;
+    private Boolean disableBranchBaseline = null;
+    private Boolean scrollAndStitchScreenshots = null;
+    private Boolean disableCORS = null;
+    private Boolean iframes = null;
+    private Map<String, Object> iframesOptions = null;
+
+    public final List<String> validOptions = Arrays.asList(
```
mykola-mokhnach, line 34: I usually wrap constants into Collections.immutableList

java/main/src/main/java/com/saucelabs/saucebindings/options/VisualOptions.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * @return instance of MutableCapabilities representing all key value pairs set in SauceOptions
+     * @see SauceSession#start()
+     */
+    public MutableCapabilities toCapabilities() {
+        String msg = "Environment Variable or System Property for 'SCREENER_API_KEY' must be provided";
+        String sauceVisualApiKey = SystemManager.get("SCREENER_API_KEY", msg);
+        capabilities.setCapability("apiKey", sauceVisualApiKey);
```
mykola-mokhnach, line 70: IMHO it is not a very secure practice to expose API keys in capabilities as their values are visible in logs. Why not, for example, use request headers for such purpose?

java/main/src/main/java/com/saucelabs/saucebindings/options/VisualOptions.java
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * @param capabilitiesToMerge a Map object representing key value pairs to convert to capabilities
+     */
+    public void mergeCapabilities(Map<String, Object> capabilitiesToMerge) {
+        for (Map.Entry<String, Object> entry : capabilitiesToMerge.entrySet()) {
+            String key = entry.getKey();
+            if (key.equals("visual")) {
+                ((Map<String, Object>) entry.getValue()).forEach(this::setCapability);
+            } else {
+                sauceOptions.setCapability(key, entry.getValue());
```
mykola-mokhnach, line 101: can sauceOptions be null?

java/main/src/test/java/com/saucelabs/saucebindings/examples/VisualOptionsTest.java
```diff
@@ -0,0 +1,41 @@
...
+        RemoteWebDriver driver = session.start();
+
+        // 6. Use the driver in your tests just like normal
+        driver.get("https://www.saucedemo.com/");
+
+        // 7. Take snapshot
+        session.takeSnapshot("Name of Snapshot");
+
+        // 8. Stop the Session
+        session.stop();
```
mykola-mokhnach, line 39: I assume this call should be in a `finally` block or AfterEach section

mykola-mokhnach: same below

java/main/src/test/java/com/saucelabs/saucebindings/integration/VisualTest.java
```diff
@@ -0,0 +1,130 @@
...
+        session.takeSnapshot("Blank");
+
+        VisualResults results = session.getVisualResults();
+
+        assertTrue(results.getPassed());
+        assertEquals("success", results.getStatus());
+
+        assertNull(results.getMessage());
+
+        assertEquals(Long.valueOf(1), results.getTotal());
```
mykola-mokhnach, line 63: `Long.valueOf(1)` could be simplified to `1L`

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
```diff
@@ -0,0 +1,85 @@
...
+        try {
+            return new URL("https://hub.screener.io/wd/hub");
+        } catch (MalformedURLException e) {
+            throw new InvalidArgumentException("Invalid URL", e);
+        }
+    }
+
+    public void newVisualTest(String testName) {
+        validateVisualStatus();
+        driver.executeScript("/*@visual.init*/", testName);
```
mykola-mokhnach, line 42: shall we first verify is the session is active/driver is initialised? I assume customers might be confused by NPE if they call this method without invoking start() first

mykola-mokhnach: same comment to the below methods. They only make sense while the session is active

java/main/src/test/java/com/saucelabs/saucebindings/integration/VisualTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+        List<VisualSnapshot> states = results.getSnapshots();
+        assertEquals(1, states.size());
+
+        VisualSnapshot snapshot = results.getSnapshots().get(0);
+        assertEquals("Blank", snapshot.getName());
+        assertEquals("VisualTest getVisualResults", snapshot.getGroupName());
+        assertEquals("accepted", snapshot.getStatus());
+        String url = snapshot.getUrl();
+        assertTrue(url.contains(GitManager.getCurrentBranch()));
```
mykola-mokhnach, line 77: I prefer to use something like `Assert.assertThat(x, CoreMatchers.containsString("foo"));`. This is giving much better error messages with actual/expected values instead of just true/false

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public void takeSnapshot(String name) {
+        validateVisualStatus();
+        driver.executeScript("/*@visual.snapshot*/", name);
+    }
+
+    public VisualResults getVisualResults() {
```
mykola-mokhnach, line 50: shall we make it possible to call this method after the session is stopped?

java/main/src/test/java/com/saucelabs/saucebindings/options/VisualOptionsTest.java
```diff
@@ -0,0 +1,269 @@
...
+    @Test
+    public void acceptsVisualOptions() {
+        SauceOptions sauceOptions = SauceOptions.firefox()
+                .setPlatformName(SaucePlatform.MAC_HIGH_SIERRA)
+                .setBrowserVersion("99")
+                .setName("Visual Options Needs a Name")
+                .build();
+
+        Map<String, Object> diffOptions = new HashMap<>();
+        diffOptions.put("structure", true);
```
mykola-mokhnach, line 52: I find this map initialisation syntax a bit too talkative. We might use some available shortcuts instead: https://www.baeldung.com/java-initialize-hashmap

mykola-mokhnach: BTW, guava is a part of selenium, so it is safe to use ImmutableMap.of()

java/main/src/test/java/com/saucelabs/saucebindings/options/VisualOptionsTest.java
```diff
@@ -0,0 +1,269 @@
...
+        assertTrue(visualOptions.getDisableBranchBaseline());
+        assertTrue(visualOptions.getScrollAndStitchScreenshots());
+        assertTrue(visualOptions.getDisableCORS());
+        assertTrue(visualOptions.getIframes());
+        assertEquals(iframesOptions, visualOptions.getIframesOptions());
+    }
+
+    @SneakyThrows
+    public Map<String, Object> serialize(String key) {
+        InputStream input = new FileInputStream(new File("src/test/java/com/saucelabs/saucebindings/options.yml"));
```
mykola-mokhnach, line 263: I usually use something like https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html with closeable streams to make sure they are properly closed

java/main/src/main/java/com/saucelabs/saucebindings/GitManager.java
titusfortner: If you don't specify anything, Screener uses `(default)`, presumably to make sure it isn't a real branch. Unfortunately, this prevents me from clicking a link with that URL because my IDE doesn't like that character. So I changed the default to `_default_`, also unlikely to be the actual branch name, but I can click on resulting links.

java/main/src/main/java/com/saucelabs/saucebindings/GitManager.java
titusfortner: Trying to think of a scenario where that could happen and be bad. It's going to just use the default value set at the top. I made these static methods/fields because no one should be switching git branches in the middle of their test execution.

java/main/src/main/java/com/saucelabs/saucebindings/GitManager.java
titusfortner: What do you mean? (and I don't remember what all I did to get while trying to get this to work when I wrote it over a year ago).

mykola-mokhnach: I mean whether it could be that `line` ends with a line break. Also, I would assume it would be safer to check the process exit code than the actual stdout

java/main/src/main/java/com/saucelabs/saucebindings/VisualResults.java
titusfortner: I usually only do enums for Constructor/Method parameters so users don't have to remember the "magic string" to get the code to work correctly. We could do it for the return value here, but the use case for 99% of people is seeing the result displayed in stdout at the end of their test. I can't see anyone asserting on this or using it in logic anywhere.

java/main/src/main/java/com/saucelabs/saucebindings/GitManager.java
mykola-mokhnach: my scenario is:
- get current branch
- the branch is changed
- get current branch and get an exception
- the previous (incorrect) branch name is returned

If we want to always return the cached value then this method should probably have an explicit condition for that

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
titusfortner: Good point. the `getSauceUrl()` in the superclass is based on `sauceUrl` field, which is a Data Center Enum, which doesn't apply for visual. How we use it here is probably going to depend on if we're planning on having a new visual data center or somehow merge the screener.io into saucelabs.com. I'll get more info.

titusfortner: Sauce does have plans to add data centers (internal link- https://saucedev.atlassian.net/browse/VSLT-2178) I'll figure out what the plan is for formatting, etc.

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
titusfortner: We want users to be able to know how to see Sauce info by looking at their Jenkins console logs, etc.
For instance, the Jenkins plugin does this by default.
Making it configurable is possible, but out of scope for current purposes.

titusfortner: Really, the link to the build is the most important thing here, imo, which is why it is frustrating that Screener doesn't expose it if the test passes. Trying to get that changed.

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
titusfortner: You know, this is a good point. It comes from our using — https://docs.saucelabs.com/basics/test-config-annotation/test-annotation/#methods which is literally *all magic strings.

When we update the implementation to use the API instead of JS to record pass/fail, then the magic strings are completely useless... Yeah, now I want to deprecate the method in `SauceSession` instead of supporting it in `VisualSession`.

mykola-mokhnach:
It is a good practice to document public methods in user-facing libs, not sure how much this principle applies to this framework

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
titusfortner: Heh, I was just getting chastised for using a fluent pattern in code I wrote for Selenium, and that was for a builder construction where it makes more sense to me. In this case it doesn't matter, there is nothing else for the Session to do after this method is called, the next line will be using the driver.

java/main/src/main/java/com/saucelabs/saucebindings/options/VisualOptions.java
titusfortner: Screener doesn't accept it in the header (because that's for Sauce credentials). It has to be in the capabilities.
As for body vs header, it might be pedantic, but it is security by convention rather than an actual difference in security. Nothing prevents a logger from recording header information even if it is less common. Either way, Sauce removes references to these keys in our logs.

java/main/src/main/java/com/saucelabs/saucebindings/options/VisualOptions.java
titusfortner: No, there are 2 constructors. One that accepts a `SauceOptions` instance, the other creates one.

java/main/src/test/java/com/saucelabs/saucebindings/examples/VisualOptionsTest.java
titusfortner: The `examples` directory is for displaying on our website, and are designed to show how easy it is to get the functionality. Best practices would be to use it with the test runner library and not have to think about it at all. :)

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
titusfortner: It needs the driver to call executeScript, and we don't have the driver after the session is stopped

java/main/src/test/java/com/saucelabs/saucebindings/options/VisualOptionsTest.java
titusfortner: lol, yeah, I just started using it. This code is a copy/paste from other stuff I've done.  I'll take a look at those examples.

titusfortner:
Thanks @mykola-mokhnach ! Yes, this definitely needs the javadocs updated before it can be released. That's just never the fun part. :)

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
mykola-mokhnach: yes I know. Currently NPE is going to be thrown in such case (which is anyway not very useful for a client). My though was that we could automatically cache that result if `stop` is called, so we don't need to query driver instance anymore.

java/main/src/main/java/com/saucelabs/saucebindings/GitManager.java
titusfortner: I don't think you can programmatically change branches in the middle of a test run. If it was found before, use that, if it wasn't found before, use default. This is a weird nice-to-have-but-not-very-important feature in the scheme of this project.

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
titusfortner: I've been treating `stop()` as the final everything is done method, but yeah, there's no reason we couldn't cache this.

java/main/src/main/java/com/saucelabs/saucebindings/GitManager.java
mykola-mokhnach: this totally makes sense to me. What I want to say is that this method is going to still spend time for the command line execution even though the branch name has been already retrieved. the condition I was talking about:

```
if (!Objects.equal(currentBranch, DEFAULT_BRANCH)) {
  return currentBranch;
}

... do existing stuff to get the branch name from git
```

titusfortner:
And looks like there are Snapshot options that I'm not supporting yet — https://docs.saucelabs.com/visual/e2e-testing/commands-options/#arguments-1

java/main/src/main/java/com/saucelabs/saucebindings/VisualSession.java
titusfortner: This is what I just got about future data center URL, so I should definitely create a `VisualDataCenter` enum for these and update when necessary.

> [11:19 AM] the future endpoints are going to be [visual.eu-central-1.saucelabs.com](http://visual.eu-central-1.saucelabs.com/) and [visual.hub.eu-central-1.saucelabs.com](http://visual.hub.eu-central-1.saucelabs.com/)

titusfortner:
Inadvertently closed because I deleted the merged branch that this was PR'd against. Trying to fix now.

PR #285, opened 2022-03-15
Java annotations

titusfortner:
Implementing these: https://docs.saucelabs.com/basics/test-config-annotation/test-annotation/

Naming stuff is hard. These sound good?
* I changed "context" to "annotate" could also be "comment"
* I changed "break" to "pause" could also be "debug"
* ~I changed "public" to be set with "setVisibility" could also be "setJobVisibility"~
*  I changed "public" to be set with "changeTestVisibility" instead of the above based on feedback
* ~I changed "job-name" to be set with "changeName" could also be "setName" or "setJobName"~
* I changed "job-name" to be set with "changeTestName" instead of the above based on feedback

This is also adding a check to make sure session has started before executing session methods.
And ensures starting/stopping network only works for Macs

This is in Draft status while I implement Python. No reason to wait on Feedback for the Java, though.

### DIFF
--- a/java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
@@ -12,6 +12,7 @@ import org.openqa.selenium.remote.RemoteWebDriver;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.List;
 
 public class SauceSession {
     @Getter protected RemoteWebDriver driver;
@@ -101,6 +102,7 @@ public class SauceSession {
      * @return an object with the accessibility analysis
      */
     public Results getAccessibilityResults(AxeBuilder builder) {
+        validateSessionStarted("getAccessibilityResults()");
         return builder.analyze(driver);
     }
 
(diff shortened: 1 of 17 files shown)


java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
```diff
@@ -118,6 +121,121 @@ public void stop(Boolean passed) {
...
+     * This is typically set in SauceOptions before starting the Session.
+     * It might be useful to set dynamically in some situations (e.g., making failing tests more visible
+     *
+     * TODO: Implement this with API because this is not currently working
+     *
+     * @param visibility who can see the job details in the Sauce Labs UI
+     * @see <a href="https://docs.saucelabs.com/basics/test-config-annotation/test-annotation/#methods">Test Annotation Methods</a>
+     * @see BaseConfigurations#setJobVisibility(JobVisibility)
+     */
+    public void setVisibility(JobVisibility visibility) {
```
nadvolod: I never liked the term "job", it's a `test`. Can it be `setTestVisibility()`?

java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
```diff
@@ -118,6 +121,121 @@ public void stop(Boolean passed) {
...
+
+    /**
+     * This is typically set in SauceOptions before starting the Session.
+     * This allows dynamially updating if necessary (e.g. name based on discovered state).
+     *
+     * @param name the test name
+     * @see <a href="https://docs.saucelabs.com/basics/test-config-annotation/test-annotation/#methods">Test Annotation Methods</a>
+     * @see BaseConfigurations#setName(String)
+     */
+    public void changeName(String name) {
```
nadvolod: Same here, `changeTestName()`?

nadvolod (commented):
Done with a few minor comments

titusfortner:
Added Python code; tests are passing; just looking for sign-off

titusfortner:
I take it back, I need to address codacy issues and update javadoc, etc
Also should add a new examples page and add to website.

titusfortner:
ok, I've fixed all the codacy I'm going to fix

titusfortner:
Ok, I removed changing visibility since it doesn't work, and it's dumb to add a method for something that doesn't work. Once the tests are passing I'm merging this.

java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
titusfortner: oops forgot to change this last time. just did.

java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
titusfortner: done

PR #290, opened 2022-04-01
Java visual update to remove VisualResults

nadvolod:
Removing `VisualResults` from Java.
Updated the API to make `projectName` and `testName` mandatory in all of the appropriate constructors.
Add advanced data driven visual example.


nadvolod:
Need to add examples like here https://github.com/saucelabs-training/demo-java/blob/main/selenium-junit4-examples/src/test/java/com/saucedemo/selenium/junit4/SimpleVisualE2ETest.java#L115-L158
Also, should create some docs on how to make the transition from old way to new way

PR #300, opened 2023-05-04
Add tunnelOwner option for python

brice:
I need this option for a project where I want to use a tunnel that I don't own. I added it for Python because I use Robot Framework.


waggledans:
implemented in https://github.com/saucelabs/sauce_bindings/pull/311

PR #302, opened 2023-09-29
[java] convert tests from JUnit 4 to JUnit 5

titusfortner:
Main tests should be using Junit 5 with the updated parallelization strategy


java/main/src/test/java/com/saucelabs/saucebindings/SauceSessionTest.java
```diff
@@ -61,29 +60,29 @@ public void sauceSessionUsesProvidedSauceConfigs() {
...
     public void defaultsToUSWestDataCenter() {
-        String expectedDataCenterEndpoint = DataCenter.US_WEST.getValue();
-        Assert.assertEquals(expectedDataCenterEndpoint, sauceSession.getDataCenter().getValue());
+        String expectedDataCenterEndpoint = DataCenter.US_WEST.getEndpoint();
+        Assertions.assertEquals(expectedDataCenterEndpoint, sauceSession.getDataCenter().getEndpoint());
     }
 
     @Test
     public void setsDataCenter() {
-        String expectedDataCenterEndpoint = DataCenter.US_EAST.getValue();
```
mmerrell, line 77: I'm not sure we should keep the East Coast data center references, should we?

mmerrell (approved):
Seems pretty straightforward to me. I don't know that we should keep the east coast data center references, but everything else looks fine.

PR #303, opened 2023-10-19
Use Google formatting and add linter to CI

titusfortner:
Move to use the Google Java Formatter.
IntelliJ plugin - https://plugins.jetbrains.com/plugin/8527-google-java-format
added github action for it as well.

This is what Selenium is using and it's so convenient for my IDE to auto format things without my needing to check codacy, I'm moving all my Java code to use it. :)


mmerrell (approved):
I see the comments in yaml, and the formatting diffs look fine to me. I think the rest of the changes are just formatting, correct? Is it an IDEA setting for running locally?

titusfortner:
not sure if you use the intelliJ auto-reformat feature (option-command-L), which is found in Code Menu. If you enable the google-java-format plugin, it will format based on Google's rules. If you miss something, the CI will now complain about it.

PR #304, opened 2023-10-24
[junit5] implement this library with extension instead of superclass

titusfortner:
I still need to tidy up the examples, but I want to show off status. Registering an extension instead of a super class is going to be significantly more flexible for toggling between local and remote, etc.

~Update: See updated code in comment below:~
Updated Update: I've re-instated original behavior and removed the builder pattern: See most recent comment

Here is the minimal implementation:

```java
@RegisterExtension static SauceBindingsExtension sauceExtension = new SauceBindingsExtension();

@Test
public void createSession() {
  sauceExtension.getSession().annotate("Running a test here");
  sauceExtension.getDriver().get("https://www.saucedemo.com/");
}
```

Tweaking options/data center before startup goes into the extension constructor:
```java
static SauceBindingsExtension sauceExtension =
  new SauceBindingsExtension(getSauceOptions(), DataCenter.US_WEST);
```
or
```diff

static SauceBindingsExtension sauceExtension =
   new SauceBindingsExtension(getCapabilities(), DataCenter.US_WEST);
```

So existing users just need to move their capability generation out of their before hook, into a static method and then pass it into the extension constructor.


titusfortner:
Ok, because I want to allow setting:
1. Capabilities
2. SauceOptions
3. DataCenter
4. Enabled/Disalbed

It does not make sense to do this all as a bunch of different constructors, so it is now set to use a Build pattern. Which, I don't love, but. makes the most sense here.

This example shows off everything:
https://github.com/saucelabs/sauce_bindings/blob/3eed57e462bd1e0b094678fe3248dfa10ea7e2ff/java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/LocalExample.java

1. Register multiple extensions (you can see there is access to the driver object in LocalTestWatcher even when used with `SauceBindingsExtension`
2. Toggles whether `SauceBindingsExtension` is enabled based on System property (`Boolean.getBoolean("sauce.enabled")`)
3. This shows `createSauceOptions()` but could just as easily be `createCapabilities()`
4. It might be a little clunky to do this every time session information is needed:
```
    if (useSauce()) {
      session.annotate("Navigating to Swag Labs");
    }
```
Would it make actual sense to have a "dummy" SauceSession object that was just a noop for every method? 🤔

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/LocalExample.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public static Boolean useSauce() {
+    System.setProperty("sauce.enabled", "true");
+    return Boolean.getBoolean("sauce.enabled");
+  }
+
+  @RegisterExtension
+  static SauceBindingsExtension sauceExtension =
+      SauceBindingsExtension.builder()
+          .capabilities(createSauceOptions())
```
nadvolod: @titusfortner this line is not very intuitive `.capabilities(createSauceOptions())`. 
After seeing the example I understand what's going on. However, will someone be able to figure out how to construct the extension without an example? There are a lot of moving objects here.

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/LocalExample.java
```diff
@@ -0,0 +1,79 @@
...
+          .enabled(useSauce())
```
nadvolod: Do we have to call `.enabled()`? What happens if we don't? What if we didn't need to call this method and instead could enable/disable an extension by setting an environment variable. This variable will be read on our end and the appropriate behavior will be enabled. This approach is more user-friendly but could also lead to some confusion about how the Sauce extension is getting enabled/disabled

nadvolod (commented):
I really like it for the most part. Just a few tiny clarifications.

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/LocalExample.java
titusfortner: I really don't want 2 different methods here. Maybe it should be `setOptions()` since people are most likely going to be sending a browser Options instance (like `ChromeOptions`) or a `SauceOptions` instance. The code signatures themselves should be sufficient?

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/LocalExample.java
titusfortner: This is to allow people to toggle local or remote execution in the same codebase. The idea is to pass in a boolean value to determine whether to use the Sauce extension or to use their local execution code. Maybe we rename it `useSauce()` instead of `enabled()`?

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/LocalExample.java
titusfortner: I've removed the builder pattern and put things back in the constructor. So no separate method, just different constructors.

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/LocalExample.java
titusfortner: Ok, it is enabled by default. It can be disabled by:
1. setting environment variable "SAUCE_DISABLED" to "true"
2. setting System Property "sauce.disabled" to "true"
3. Commenting out the registration lines 😂

titusfortner:
Updated again:
1. Enabled by default because that's what Registering an Extension should do
2. Disable by one of:
    * Set Environment Variable of `"SAUCE_DISABLED"` to `"true"`
    * Set System Property of `"sauce.disabled"` to `"true"`
    * Comment out the Registration line 😂 
3. Back to Constructors instead of Builder Pattern because more straightforward
4. There's one chunk of code that we can remove when #305 is merged/released

titusfortner:
Hmm, not sure why local example isn't passing, since it should be exactly the same as disabled test. They work locally and that error indicates the driver and browser versions do not match. I'll turn on logging to see what's different.

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/LocalExample.java
```diff
@@ -0,0 +1,87 @@
...
+  WebDriver driver;
+  SauceSession session;
+
+  @RegisterExtension
+  static SauceBindingsExtension sauceExtension = new SauceBindingsExtension(getSauceOptions());
+
+  @RegisterExtension TestWatcher testWatcher = new LocalTestWatcher();
+
+  @BeforeAll
+  public static void disableSauce() {
```
nadvolod: @titusfortner I don't understand why we are disabling Sauce in a `beforeAll`?

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/OptionsExample.java
```diff
@@ -0,0 +1,42 @@
...
+import com.saucelabs.saucebindings.junit5.SauceBindingsExtension;
+import com.saucelabs.saucebindings.options.SauceOptions;
+import java.time.Duration;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.chrome.ChromeOptions;
+
+public class OptionsExample {
```
nadvolod, line 14: @titusfortner one of the reasons why we started Sauce Bindings was to increase the amount of code needed to get started. I wonder how much less code this is vs the legacy way?

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/LocalExample.java
titusfortner: To show how to run locally. Running on sauce is default. Need to disable to run local.

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/OptionsExample.java
titusfortner: Quite a bit still. I'll add examples without the extension to show it.
But look at quick start example.

titusfortner: So, the big difference is the Quick Start; you get all the defaults without any boilerplate:
[with extension](https://github.com/saucelabs/sauce_bindings/blob/c67d25f33578b4151b26295ec96ea83734805524/java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/QuickStartExample.java) (27 lines) vs [without extension](https://github.com/saucelabs/sauce_bindings/blob/c67d25f33578b4151b26295ec96ea83734805524/java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/without/QuickStartExample.java) (93 lines)

But most users are going to have custom capabilities and want to toggle between running on Sauce & running locally. So the "real world" example comparison looks more like this:
[with extension](https://github.com/saucelabs/sauce_bindings/blob/c67d25f33578b4151b26295ec96ea83734805524/java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/ToggleLocalExample.java) (88 lines) vs [without extension](https://github.com/saucelabs/sauce_bindings/blob/c67d25f33578b4151b26295ec96ea83734805524/java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/without/ToggleLocalExample.java) (155 lines)

nadvolod (approved):

nadvolod:
@titusfortner you want me to merge this?

titusfortner:
Let me figure out why it is failing the test first. 😦

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
```diff
@@ -0,0 +1,112 @@
...
+          .filter(line -> !line.contains("sun"))
+          .forEach(session::annotate);
+
+      session.stop(false);
+    }
+  }
+
+  // TODO: Implement this in SauceSession directly
+  private boolean isExtensionDisabled() {
+    String value = System.getenv("SAUCE_DISABLED");
```
diemol, line 111: This would be better as a system property so the user can set it in code or via the execution parameters.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
```diff
@@ -0,0 +1,112 @@
...
+    if (isExtensionDisabled()) {
+      return;
+    }
+
+    try {
+      session.stop(true);
+    } catch (NoSuchSessionException e) {
+      System.out.println(
+          "Driver quit prematurely; Remove calls to `driver.quit()` to allow SauceBindingsExtension"
+              + " to stop the test");
```
diemol: We should use a proper logger here.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
```diff
@@ -0,0 +1,112 @@
...
+import org.junit.jupiter.api.extension.TestWatcher;
+import org.openqa.selenium.Capabilities;
+import org.openqa.selenium.NoSuchSessionException;
+import org.openqa.selenium.WebDriver;
+
+public class SauceBindingsExtension implements TestWatcher, BeforeEachCallback {
+  private final SauceOptions sauceOptions;
+  private final DataCenter dataCenter;
+  @Getter private SauceSession session;
+  @Getter private WebDriver driver;
```
diemol, line 24: Are these the two only uses of Lombok? If so, let's remove the dependency and write a getter/setter.

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/OptionsExample.java
diemol: I think this looks fine.

java/junit5/pom.xml
```diff
@@ -61,6 +61,12 @@
             <artifactId>sauce_bindings</artifactId>
             <version>1.4.0</version>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.28</version>
+            <scope>provided</scope>
+        </dependency>
```
diemol, line 69: I only saw usage in two parts. Could we remove this dependency?

diemol (changes requested):
Just some minor comments that we could address quickly.

Thank you, @titusfortner!

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: We're using it much more extensively in main Sauce Bindings, so we're still using it in this project quite a bit.

java/junit5/pom.xml
titusfortner: If we're going to use it in one part of the monorepo, I think the methods should be available everywhere

diemol: Fair enough.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
diemol: 👍

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
diemol: Replaced.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
diemol: Implemented.

diemol (approved):

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
```diff
@@ -106,7 +106,7 @@ public void testFailed(ExtensionContext context, Throwable cause) {
 
   // TODO: Implement this in SauceSession directly
   private boolean isExtensionDisabled() {
-    String value = System.getenv("SAUCE_DISABLED");
+    String value = System.getProperty("SAUCE_DISABLED");
```
titusfortner: Our implementation of `getenv()` method checks both property and environment variable

titusfortner: oh, maybe that was something special in SauceBindings not this library

PR #306, opened 2023-11-30
[java] all user to no-op Sauce Session methods

titusfortner:
Sometimes you want to run your tests locally and sometimes you want to run on Sauce Labs.

If you add a bunch of Sauce Labs specific features to your code (which we want to encourage), it's going to be a pain to disable or add conditionals for everything. If we allow users to disable running on Sauce with a System Property or Environment Variable, then they can keep all those custom methods (like annotations or stop methods) where they are without conditionals.

For instance, the JUnit5 extension will do this:
```java
  @Override
  public void testFailed(ExtensionContext context, Throwable cause) {
    if (isExtensionDisabled()) {
      return;
    }

    if (session != null) {
      session.annotate("Failure Reason: " + cause.getMessage());
      Arrays.stream(cause.getStackTrace())
          .map(StackTraceElement::toString)
          .filter(line -> !line.contains("sun"))
          .forEach(session::annotate);
      session.stop(false);
    }
  }
```

With this PR, we can remove the conditional to return. `stop()` and `annotate()` just won't do anything if Sauce is disabled.

This is easy with `void` methods. Just don't do anything.
`start()` and the accessibility methods actually return objects, though. I'm not sure whether to error or return `null`. I kind of want to return `null` for `start()` and throw an error for accessibility, but that doesn't seem right for them to be different.

What do you think? Does this make sense, or does it seem counter-intuitive? Should we return nulls or throw an error?

### DIFF
--- a/java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
@@ -42,7 +42,8 @@ public class SauceSession {
    * Checks if the feature is disabled in either system property or environment variable.
    *
    * @return false if the feature is disabled, true otherwise.
-   */public boolean isEnabled() {
+   */
+  public boolean isEnabled() {
     return !SystemManager.getBoolean("sauce.disabled")
         && !SystemManager.getBoolean("SAUCE_DISABLED");
   }
@@ -54,7 +55,7 @@ public class SauceSession {
    */
   public RemoteWebDriver start() {
     if (!isEnabled()) {
-      return null;
+      throw new RuntimeException("Sauce is Disabled by System Property or Environment Variable.");
     }
 
     this.driver = createRemoteWebDriver(getSauceUrl(), sauceOptions.toCapabilities());
--- a/java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
@@ -39,11 +39,24 @@ public class SauceSession {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


titusfortner:
Hmm, yeah, we don't want to return `null`, we want to raise exceptions, so for these, users will need to:

```java
  session = sauceExtension.getSession();
  driver = session.isEnabled() ? sauceExtension.getDriver() : getLocalDriver();
```

java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
```diff
@@ -231,6 +282,10 @@ public void changeTestName(String name) {
    * @see BaseConfigurations#setTags(List)
    */
   public void addTags(List<String> tags) {
+    if (!isEnabled()) {
```
nadvolod, line 285: @titusfortner all these conditionals in our methods seem kind of annoying. Is there any way to just turn off the method actions if the extension `!isEnabled()`. A single place to turn on/off the operations, like a fuse box.
However, if we don't have these conditionals then that means that the user has to think about them, right?

titusfortner: Right, the idea is to put the ugliness in our code instead of the user needing to put it in theirs. There's definitely a more clever way to delegate the functionality, but I think it would be overkill and also possibly less obvious. We can change the implementation itself later if we need.

diemol (commented):
Seems that `validateSessionStarted` is a common place where most Sauce session methods go. You could move that logic there, and also put it in the `start()` and `stop()` methods?

titusfortner:
This was implemented in #318

PR #318, opened 2024-08-23
[java] implement global toggle to noop session commands

titusfortner:
# One-line summary

This allows users to put things like annotations and other sauce bindings specific code in frameworks and tests without needing to put conditionals everywhere

## Description
We need to make it as easy as possible for people to toggle between running on Sauce and running locally. Obviously there is a limit to how much can be done

## Types of Changes

## Considerations

* Is this the right system property name?
* Should we have a public setters/getters for this on Session class?
* I'm not positive this works the way we need it to in the test runner code, but I can't test it without building it.


nadvolod (approved):

titusfortner:
@nadvolod, fyi, I disabled the merge commit button for this repo. The best way to keep the log clean is to do a squash merge on PRs. That allows for a bunch of "tidy" type commits without needing to go into the permanent log.

PR #320, opened 2024-09-16
[java] update property name

titusfortner:
# One-line summary

Rename toggle property to be more descriptive

## Description
To toggle off Sauce code, set `sauce.disabled` to `"true"` rather than using `saucelabs` to `"false"`

## Types of Changes

- Refactor/improvements

## Updates:

* Allow disabling with environment variable as well as system property
* Rename Examples from `*Test.java` to `*Example.java`
* Run Examples sequentially after Tests to keep system property from propagating to other tests

### DIFF
--- a/java/main/pom.xml
+++ b/java/main/pom.xml
@@ -125,17 +125,41 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <version>3.1.2</version>
-        <configuration>
-          <properties>
-            <configurationParameters>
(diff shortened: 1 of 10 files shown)


diemol (commented):
I see some tests failing.

nadvolod:
@titusfortner do we need an acceptance test that validates that this behavior works as expected? I remember that you had concerns. I also see a bunch of `//TODO` in the `ToggleLocalExample.java`

nadvolod (approved):

diemol (approved):

PR #330, opened 2024-09-27
Adding tags for Sauce Bindings, and restructuring tag setters so they…

mmerrell:
… append, not overwrite

# One-line summary

> Issue : #1234 (only if appropriate)

## Description
We want to be able to track who is using Sauce Bindings and overall usage. This PR adds that functionality, and makes it so that when you set tags, you append to the existing list, rather than overwriting it.

## Types of Changes
_What types of changes does your code introduce? Keep the ones that apply:_

## Tasks
_List of tasks you will do to complete the PR_
  - [ ] Created Task 1
  - [ ] Created Task 2
  - [ ] To-do Task 3

## Review
_List of tasks the reviewer must do to review the PR_
- [x] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes
These should highlight any db migrations, feature toggles, etc.

### DIFF
--- a/java/main/src/main/java/com/saucelabs/saucebindings/options/SauceLabsOptions.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/options/SauceLabsOptions.java
@@ -7,7 +7,6 @@ import com.saucelabs.saucebindings.SauceSession;
 import com.saucelabs.saucebindings.SystemManager;
 import java.net.URL;
 import java.util.*;
-
 import lombok.Getter;
 import lombok.Setter;
 import lombok.experimental.Accessors;
@@ -147,14 +146,13 @@ public class SauceLabsOptions extends BaseOptions {
     } else if ("tags".equals(key)) {
       try {
         List<String> tagsList = (List) value;
-        for (String tag: tagsList) {
+        for (String tag : tagsList) {
           tags.add(tag);
         }
       } catch (ClassCastException e) {
         throw new RuntimeException(e);
       }
-    }
-    else {
+    } else {
       super.setCapability(key, value);
     }
   }
@@ -168,8 +166,8 @@ public class SauceLabsOptions extends BaseOptions {
   }
 
   /**
-   * @deprecated This method is no longer supported
    * @return whether CI is accounted for in code
+   * @deprecated This method is no longer supported
    */
   @Deprecated
   public boolean isKnownCI() {
--- a/java/main/src/main/java/com/saucelabs/saucebindings/options/SauceLabsOptions.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/options/SauceLabsOptions.java
@@ -6,10 +6,8 @@ import com.saucelabs.saucebindings.Prerun;
 import com.saucelabs.saucebindings.SauceSession;
 import com.saucelabs.saucebindings.SystemManager;
 import java.net.URL;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
+
 import lombok.Getter;
 import lombok.Setter;
 import lombok.experimental.Accessors;
@@ -44,7 +42,18 @@ public class SauceLabsOptions extends BaseOptions {
   private Boolean recordVideo = null;
   private String screenResolution;
   private String seleniumVersion;
-  private List<String> tags = null;
+
+  public BaseOptions setTags(List<String> tags) {
+    this.tags.addAll(tags);
+    return this;
+  }
+
+  public BaseOptions setTag(String tags) {
+    this.tags.add(tags);
+    return this;
+  }
+
+  private List<String> tags = new ArrayList<>(Arrays.asList("sauce-bindings", "java"));
   private String timeZone;
   private String tunnelIdentifier;
   private Boolean videoUploadOnPass = null;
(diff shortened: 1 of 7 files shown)


java/main/src/main/java/com/saucelabs/saucebindings/options/SauceLabsOptions.java
```diff
@@ -6,10 +6,7 @@
 import com.saucelabs.saucebindings.SauceSession;
 import com.saucelabs.saucebindings.SystemManager;
 import java.net.URL;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
titusfortner, line 9: convention is not to use wildcards in imports

java/main/src/main/java/com/saucelabs/saucebindings/options/SauceLabsOptions.java
```diff
@@ -44,7 +41,18 @@ public class SauceLabsOptions extends BaseOptions {
   private Boolean recordVideo = null;
   private String screenResolution;
   private String seleniumVersion;
-  private List<String> tags = null;
+
+  public BaseOptions setTags(List<String> tags) {
+    this.tags.addAll(tags);
```
titusfortner, line 46: What breaks if we do not change the existing behavior for overwriting all tags?

java/main/src/main/java/com/saucelabs/saucebindings/options/SauceLabsOptions.java
```diff
@@ -135,6 +143,15 @@ protected void setCapability(String key, Object value) {
                 prerunMap.put(Prerun.valueOf(keyString), val);
               });
       setPrerun(prerunMap);
+    } else if ("tags".equals(key)) {
```
titusfortner, line 146: Does this need special handling? If setTag() is adding it to current tags field, then doesn't the default getTags() return what we need?

java/main/src/test/java/com/saucelabs/saucebindings/options/ChromeConfigurationsTest.java
```diff
@@ -2,10 +2,7 @@
 
 import com.saucelabs.saucebindings.*;
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
titusfortner, line 5: convention is not to use wildcards in imports

java/main/src/test/java/com/saucelabs/saucebindings/options/EdgeConfigurationsTest.java
```diff
@@ -2,10 +2,7 @@
 
 import com.saucelabs.saucebindings.*;
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
titusfortner, line 5: convention is not to use wildcards in imports

java/main/src/test/java/com/saucelabs/saucebindings/options/FirefoxConfigurationsTest.java
```diff
@@ -2,10 +2,7 @@
 
 import com.saucelabs.saucebindings.*;
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
titusfortner, line 5: convention is not to use wildcards in imports

java/main/src/test/java/com/saucelabs/saucebindings/options/InternetExplorerConfigurationsTest.java
```diff
@@ -2,10 +2,7 @@
 
 import com.saucelabs.saucebindings.*;
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
titusfortner, line 5: convention is not to use wildcards in imports

java/main/src/test/java/com/saucelabs/saucebindings/options/SafariConfigurationsTest.java
```diff
@@ -2,10 +2,7 @@
 
 import com.saucelabs.saucebindings.*;
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
titusfortner, line 5: convention is not to use wildcards in imports

java/main/src/test/java/com/saucelabs/saucebindings/options/SauceOptionsTest.java
```diff
@@ -4,10 +4,7 @@
 import java.io.FileInputStream;
 import java.io.InputStream;
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
titusfortner, line 7: convention is not to use wildcards in imports

titusfortner (changes requested):
So, the main wrench I want to throw into this is.... I think this information should get stored in custom data and not as a tag. For instance, i also want to note what CI tool is getting used, and I want us to differentiate between JUnit5 and JUnit4 and TestNG, etc as well

titusfortner:
@mmerrell interesting, I verified that we can filter on custom data field in Looker. I've also seen that we already have clients that are copying their failure messages into custom data fields at the end of the run, which is interesting.

titusfortner:
Thanks for the work on this, but I'm closing this in favor of #334 since we'd prefer multiple key/value pairs with versions in customData instead of the tag implementation that we originally discussed.

PR #331, opened 2024-10-02
[testng] implement as a test listener instead of requiring inheritance

titusfortner:
# One-line summary

implement as a test listener instead of requiring inheritance

## Description
This is ~the same as~ similar to what we did for JUnit 4 & JUnit 5; unlike the others I couldn't make the startup happen automatically before the test's before hook (`@BeforeMethod`) executes. So, you have to explicitly tell the listener class to start things, and because the values are stored in the test context you have to explicitly pull them out to use them.

***Update***: Turns out Context was not completely Thread Safe, so I had to re-re do it the way JUnit 5 was redone storing sessions as a unique key in the context. Created a separate SessionContext to manage everything since we can't use a constructor for the Listener and it stopped making sense to put all the unrelated things into that listener class.

Reasoning:
We need to make it easier for people to add this into their existing code
That means we don't want to force a specific inheritance on the user
Using a Listener instead of a SuperClass should make this more extensible/compatible

## Types of Changes

## Considerations
This should be easier to work with, but TestNG gets complicated quickly with XML files and ThreadLocals
The implementation here is using TestContext to store the session and driver variables to get around Threading limitations
I even added a parameterized example as proof of concept.
Ideally this is included as part of Sauce Bindings 2.0
I need to make sure toggling local/remote works


titusfortner:
Marking as draft while I verify that this is doing what it needs to in the larger test suite.

titusfortner:
Got this verified working. Needed to add ThreadLocal still in the test parts. Need to either always pull driver and session from context or use ThreadLocal, and the latter is more common.

diemol (approved):

PR #332, opened 2024-10-04
[junit5] fix bug causing all tests to be named the same

titusfortner:
# One-line summary
The extension uses the same options instance for all tests, and users cannot change options after initialization, so the current check does not make sense; user can use session.changeTestName() if needed

## Types of Changes
- Bug fix (non-breaking change which fixes an issue)


titusfortner:
fixed in #334

PR #333, opened 2024-10-04
[junit5] update pom so it gets both the jupiter engine and api artifacts

titusfortner:
# One-line summary

Got an error trying to use this library in another project because it needed the api jar as well

## Types of Changes
- Bug fix (non-breaking change which fixes an issue)


titusfortner:
part of #334

PR #334, opened 2024-10-10
Preparation for 2.0 Release

titusfortner:
It's really hard to break things up into small PRs when a bunch of things have to change, especially if we can't do timely reviews on even one-liners.

So here is a 2,000 liner.

* Fixes thread safety issue with JUnit 5
* Removes deprecated code and sets snapshot to 2.0 beta
* implements serialization so SauceOptions can be copied
* implements abort method to quit session without updating test result for when tests are aborted
* Fixed up CITools class so you can get the CI tool name and the logic is more obvious
* Add CI Tool and Sauce Bindings to Custom Data on startup
* Bunch of fixes for testing when disabling Sauce in properties

## To Do
* Ensure TestNG is thread safe (I think it is probably good, but need to verify it doesn't need `copy()`)
* JUnit 4 is definitely not thread safe
* JUnit 4 and TestNG plugins both need to update to add things to custom data


java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
```diff
@@ -1,114 +1,171 @@
...
+    getStore(context).put("session", session);
+    getStore(context).put("driver", driver);
   }
 
-  @Override
-  public void beforeEach(ExtensionContext context) {
-    if (isExtensionDisabled()) {
-      return;
+  private SauceOptions updateOptions(ExtensionContext context) {
+    SauceOptions options = sauceOptions.copy();
```
diemol, line 55: Why is this needed? Does JUnit have a single instance of `SauceBindingExtention`?

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
```diff
@@ -1,114 +1,171 @@
...
+  private ExtensionContext.Store getStore(ExtensionContext context) {
+    return context.getStore(
+        ExtensionContext.Namespace.create(getClass(), context.getRequiredTestMethod()));
   }
 
   @Override
   public void testSuccessful(ExtensionContext context) {
-    if (isExtensionDisabled()) {
-      return;
+    if (!SauceSession.isDisabled()) {
```
diemol, line 92: Why is `isDisabled()` in the `SauceSession` and not in `SauceBindingExtention`? Could you tell me what the reason was for this?

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
```diff
@@ -1,114 +1,171 @@
...
+      SauceSession session = (SauceSession) getStore(context).get("session");
```
diemol, line 93: Why is `SauceSession` not at the class level like `SauceOptions`? It seems strange to retrieve it from the extension's context. I might be misunderstanding how extensions in JUnit work.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
```diff
@@ -1,114 +1,171 @@
...
+      RemoteWebDriver driver = session.getDriver();
+      try {
+        session.stop(true);
+      } catch (NoSuchSessionException e) {
+        LOGGER.severe(
```
diemol, line 98: Why `severe`? Is this breaking the test execution? Wouldn't a `warning` work better?

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
```diff
@@ -1,114 +1,171 @@
...
+  @Override
+  public void testFailed(ExtensionContext context, Throwable cause) {
+    if (!SauceSession.isDisabled()) {
+      SauceSession session = (SauceSession) getStore(context).get("session");
+      try {
+        session.annotate("Failure Reason: " + cause.getMessage());
+
+        Arrays.stream(cause.getStackTrace())
+            .map(StackTraceElement::toString)
+            .filter(line -> !line.contains("sun"))
```
diemol, line 114: What is this `"sun"` magic string? Please add a comment to understand better what it means.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
```diff
@@ -1,114 +1,171 @@
...
     }
   }
 
-  // TODO: Implement this in SauceSession directly
-  private boolean isExtensionDisabled() {
-    String value = System.getenv("SAUCE_DISABLED");
-    return Boolean.parseBoolean(value) || Boolean.getBoolean("sauce.disabled");
+  @Override
+  public void testDisabled(ExtensionContext context, Optional<String> reason) {
+    LOGGER.info("A Sauce session was not started for " + context.getDisplayName() + " because " + reason);
```
diemol: Is this really needed? The user should know the test is disabled.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
```diff
@@ -1,114 +1,171 @@
...
-  private static final Logger logger = Logger.getLogger(SauceBindingsExtension.class.getName());
-  private final SauceOptions sauceOptions;
-  private final DataCenter dataCenter;
-  @Getter private SauceSession session;
-  @Getter private WebDriver driver;
+public class SauceBindingsExtension implements TestWatcher, BeforeEachCallback, ParameterResolver {
+  private static final Logger LOGGER = Logger.getLogger(SauceBindingsExtension.class.getName());
+  private final String buildName;
+  protected SauceOptions sauceOptions = new SauceOptions();
+  protected DataCenter dataCenter = DataCenter.US_WEST;
```
diemol: If the builder is setting this value, why is it needed here?

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
```diff
@@ -1,114 +1,171 @@
...
-            });
-
-    this.sauceOptions.mergeCapabilities(capabilitiesMap);
-    this.dataCenter = dataCenter;
-  }
+  @Override
+  public void beforeEach(ExtensionContext context) {
+    SauceOptions options = updateOptions(context);
+    SauceSession session = new SauceSession(options);
+    session.setDataCenter(dataCenter);
```
diemol, line 47: If a session was started with `EU_CENTRAL`, wouldn't this override it to `US_WEST`? Is there a test for this?

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/CapabilitiesTest.java
```diff
@@ -21,16 +24,17 @@ public static Capabilities getCapabilities() {
...
   @RegisterExtension
-  static SauceBindingsExtension sauceExtension = new SauceBindingsExtension(getCapabilities());
+  static SauceBindingsExtension sauceExtension =
+      new SauceBindingsExtension.Builder().withCapabilities(getCapabilities()).build();
 
   @Test
-  public void useCustomCapabilities() {
-    RemoteWebDriver driver = (RemoteWebDriver) sauceExtension.getDriver();
-    Assertions.assertEquals("Safari", driver.getCapabilities().getBrowserName());
+  public void useCustomCapabilities(SauceSession session, WebDriver driver) {
```
diemol, line 30: Do all users need to change the signature of their test when using the extension? Is this needed?

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/OptionsTest.java
```diff
@@ -17,16 +19,17 @@ public static SauceOptions getSauceOptions() {
   }
 
   @RegisterExtension
-  static SauceBindingsExtension sauceExtension = new SauceBindingsExtension(getSauceOptions());
+  static SauceBindingsExtension sauceExtension =
```
diemol, line 22: What is the difference between this test and `CapabilitiesTest`? It looks like a duplicate.

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/CapabilitiesTest.java
diemol: If this is needed, can we only leave `SauceSession`, and users can get the `driver` from it?

If we want SauceBindings to be used with other frameworks like Playwright, putting the `WebDriver` there would complicate things in the future.

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/ToggleLocalExample.java
```diff
@@ -14,83 +10,47 @@
...
   @BeforeAll
-  public static void disableSauce() {
-    System.setProperty("sauce.disabled", "true");
-  }
-
-  @AfterAll
-  public static void resetSauce() {
-    System.clearProperty("sauce.disabled");
+  public static void toggleSauce() {
+    System.setProperty("sauce.disabled", "false");
```
diemol, line 25: Is there a test for the `true` value? Did I miss it?

java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
```diff
@@ -35,19 +34,16 @@ public abstract class CITools {
...
+    if (ciToolName == null) {
       for (Map.Entry<String, String> tool : KNOWN_TOOLS.entrySet()) {
         if (SystemManager.get(tool.getValue()) != null) {
-          buildName = SystemManager.get(BUILD_VALUES.get(tool.getKey()).get(0));
-          buildNumber = SystemManager.get(BUILD_VALUES.get(tool.getKey()).get(1));
-          if (buildName != null && buildNumber != null) {
-            buildInfoStored = true;
-            break;
-          }
+          ciToolName = tool.getKey();
```
diemol, line 40: Why don't you return the value here?

java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
```diff
@@ -56,8 +52,11 @@ private static void storeBuildInfo() {
    * @return the constant name of the build
    */
   public static String getBuildName() {
-    storeBuildInfo();
-    return buildName;
+    if (!Objects.equals(getCiToolName(), "unknown")) {
```
diemol: Why not `equalsIgnoreCase`? This compares strings, while `equals` compares hash codes.

java/main/src/main/java/com/saucelabs/saucebindings/options/SauceOptions.java
```diff
@@ -1,6 +1,12 @@
 package com.saucelabs.saucebindings.options;
 
 import com.saucelabs.saucebindings.*;
+
```
diemol: Single imports for line 3.

java/main/src/main/java/com/saucelabs/saucebindings/options/SauceOptions.java
```diff
@@ -206,6 +212,23 @@ public Map<Timeouts, Integer> getTimeouts() {
     }
   }
 
+  public SauceOptions copy() {
```
diemol, line 220: I still miss why this is needed. Are we doing the same thing in Selenium?

java/main/src/main/java/com/saucelabs/saucebindings/options/SauceLabsOptions.java
```diff
@@ -148,20 +142,12 @@ public String getBuild() {
...
 
-  /**
-   * @deprecated This method is no longer supported
-   * @return whether CI is accounted for in code
-   */
-  @Deprecated
-  public boolean isKnownCI() {
-    return CITools.getBuildName() != null;
+  public Map<String, Object> getCustomData() {
+    customData.putIfAbsent("sauce-bindings", "java"); // might be set by plugin already
```
diemol: This needs to set the bindings version as well.

diemol (commented):
I think a test suite that runs in parallel is needed to make sure all these changes are actually doing what you are expecting.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: Yes, the SauceBindingsExtension instance is shared for the tests

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: 1. Implement it in one place instead of each of the libraries
2. It's much easier to deal with empty and null returns when disable code is in Bindings

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: Yes, this is how we maintain thread safety by ensuring that changes are limited to a specific test context.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: Because if you get this error it means the user code is preventing the library from being useful. If Sauce isn't getting pass/fail information, then they shouldn't be bothering with this library. So I don't see this as a caution but as an essential thing to be fixed.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: You're right, this implementation is dumb.
The goal is to keep the stack trace from creating 1,000 separate sauce context commands.
We should find a better way to limit to lines we know are useful.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: Not necessarily(?) tests can be disabled dynamically, and it might not be obvious looking at the console log which apply? (see https://github.com/saucelabs/sauce_bindings/pull/334/files/cde48e73218c144a1ed5acc804d063c01fe5f7a1#diff-fb8851e65411919c7b5db33a84c31f6f1ec2554e6ddc6118952da080bc55499eR17)

Whether logging is necessary because the information already shows up in normal maven execution is possible. I think we need to have something implemented there, and I chose to log a statement about how the status affects this extension. If we think it is duplicated, we can set it to FINE instead?

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: Ah, right, I unnecessarily added it in both places. I'll see which one makes more sense.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: data center isn't read until `start()` is called. But, you're right I'm not sure why we're doing both constructor and setter. I'll investigate.

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/CapabilitiesTest.java
titusfortner: Yes, but it can be in a before hook in their base class that just stores the variables. The examples are skipping that for simplicity right now.

This implementation is great because you an do either or both in either order or none.
I originally had it only for Session, and put in code for user to call `session.getDriver()`. But I think it is simpler to show it with both.

It's just adding a conditional to 2 methods to allow:
```
@BeforeEach
public void setup(SauceSession session, Playwright playwright)
```

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/OptionsTest.java
titusfortner: Options is SauceBindings `SauceOptions` and Capabilities is Selenium Capabilities (e.g., `ChromeOptions`)

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
mmerrell: We're going to need to document this if it's a change to how a user will perceive their results, or if it changes behavior they're used to. In general I agree, I just want to make sure we don't irritate people and make them switch back to non-bindings.

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples/ToggleLocalExample.java
titusfortner: yeah, I moved these around. There is a ToggleTest now that is disabled if the property isn't set, and is run by maven with a separate execution specified in the pom file.

I didn't want to add the disable annotations to the example code, but I still wanted to show how a user could use the code in the real world.

java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
titusfortner: I guess because it breaks and returns, so no performance improvement. But yeah, can replace break with return just as easily; probably more clear.

java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
titusfortner: Good call

java/main/src/main/java/com/saucelabs/saucebindings/options/SauceOptions.java
titusfortner: Selenium capabilities classes implement Serializable which means you can implement a copy feature easily if you want (which I didn't know before this week)

I tried a bunch of other things first and ran into one issue after another; I settled on this because it guarantees its not a shallow copy without my having to write dozens of tests, and it is overall a very simple implementation

java/main/src/main/java/com/saucelabs/saucebindings/options/SauceLabsOptions.java
titusfortner: Good call... So maybe like this?
```
sauce-bindings: "java"
sauce-bindings-java:  "2.0.1"
sauce-bindings-junit5: "2.0.0"
```

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/CapabilitiesTest.java
diemol: Requiring the user to change their code for every test might make people steer away.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: So 3 ways to do this:
1. Raise Exception - tell user this is absolutely not how to use the library
2. Log Severe - let the user do it but express quite strongly that it is really bad
3. Log Warning - tell users this isn't how they should be doing it

I can make an argument for why we should do 1, but I figured it would be better swallow the exception and complain loudly in the logs.

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/CapabilitiesTest.java
titusfortner: Would it be less confusing to always use a before hook?

Like it's obvious that the difference between what we had previously vs now is simple:
```
@BeforeEach
public void setUp() {
  this.session = extension.getSession();
  this.driver = extension.getDriver();
}
```

Now:
```
@BeforeEach
public void setUp(SauceSession session, WebDriver driver) {
  this.session = session;
  this.driver = driver;
}
```

Maybe I'm being too cute by removing the before hook in the examples just for fewer lines of code.

titusfortner:
> I think a test suite that runs in parallel is needed to make sure all these changes are actually doing what you are expecting.

1. This test suite does run in parallel, but it doesn't have enough tests to consistently hit the race conditions
2. I've been using another test suite to verify that it is working properly, which isn't ideal, but has been helpful.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: No, this is right. SauceSession does not take data center in the constructor, and requires changing the default to be done by a setter. i'm not sure if this is best, but also it doesn't matter if people just use the extension. :)

java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/CapabilitiesTest.java
titusfortner: Oh no, I did the right thing. The _examples_ all show it in a before hook for clarity. In the _tests_ I'm using a shortcut because it's the tests.

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: Fixed: https://github.com/saucelabs/sauce_bindings/pull/334/commits/b67dc5b7cde32312cbcb73fdbc2b0c028fc7ecc8

java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
titusfortner: https://github.com/saucelabs/sauce_bindings/pull/334/commits/5716618f7e2924c3851e61263341071309f0880c

java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
titusfortner: https://github.com/saucelabs/sauce_bindings/pull/334/commits/739a24eb03b71261fcd7608f2fb3ec70ce7541af

java/main/src/main/java/com/saucelabs/saucebindings/options/SauceLabsOptions.java
titusfortner: https://github.com/saucelabs/sauce_bindings/pull/334/commits/c8b3565b0133c72108f297d14061ac9238c64168 & https://github.com/saucelabs/sauce_bindings/pull/334/commits/648c7f1eee94a7d3dd44fb79ac41d0018d9cbc43

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: So when it isn't disabled, it looks like:
```
[INFO] Running com.saucelabs.saucebindings.junit5.ToggleTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.276 s -- in com.saucelabs.saucebindings.junit5.ToggleTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
``` 

When it is disabled, without the output is:
```
[INFO] Running com.saucelabs.saucebindings.junit5.ToggleTest
Oct 11, 2024 12:43:39 PM com.saucelabs.saucebindings.junit5.SauceBindingsExtension testDisabled
[WARNING] Tests run: 1, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 0.075 s -- in com.saucelabs.saucebindings.junit5.ToggleTest
[INFO] 
[INFO] Results:
[INFO] 
[WARNING] Tests run: 1, Failures: 0, Errors: 0, Skipped: 1
```
It doesn't say what was disabled or why. 
I think there is value in noting that tests were disabled because of that property getting set and which they are:
```
INFO: A Sauce session was not started for disableSauce(SauceSession, WebDriver) because Optional[System property [sauce.disabled] does not exist]
```

java/junit5/src/main/java/com/saucelabs/saucebindings/junit5/SauceBindingsExtension.java
titusfortner: I'm just going to throw everything into one line and we can improve it later - https://github.com/saucelabs/sauce_bindings/pull/334/commits/8222c43c0fdbea73230a2873dc9df75d50252246

diemol (approved):

nadvolod:
> > I think a test suite that runs in parallel is needed to make sure all these changes are actually doing what you are expecting.
> 
> 1. This test suite does run in parallel, but it doesn't have enough tests to consistently hit the race conditions
> 2. I've been using another test suite to verify that it is working properly, which isn't ideal, but has been helpful.

I agree with Diego to avoid future regressions. How many tests do we need in parallel to hit race conditions? How hard is it to add a TestClass with enough dumb parallel tests to invoke the race conditions? @titusfortner

titusfortner:
I've got a test suite I'm using, we can discuss if it should get added to this repo or not.

titusfortner:
for future reference: the force pushes above was me playing around with adding fixes for CI tooling things, but I'm going to make that another PR.

PR #335, opened 2024-10-16
[java] create CI Tool classes instead of using a map of constants

titusfortner:
# One-line summary

Allow user-defined build name and number as well as better GitHub default

## Description
Instead of a complicated map of environment variables, this creates a new class for each CI Tool.

As for GitHub, `GITHUB_JOB` for this repo is e.g., `main`, `junit5`, `junit4` (https://github.com/saucelabs/sauce_bindings/blob/main/.github/workflows/java.yml#L62), which isn't super helpful on Sauce —  `Automated / Builds/ main: 149` (https://app.saucelabs.com/builds/vdc/c9e1a2beb1b03686b0c248e4dc73921b)

This PR adds `GITHUB_WORKFLOW`, so it will look like: `Java / main: 149`. It might be even better to add project? so `saucebindings / Java / main: 149`? 

Also, user can now adjust it to be whatever they want:
```
System.setProperty("sauce.build.name", System.getenv("GITHUB_WORKFLOW") + " -  " +  System.getenv("GITHUB_JOB"))
```

`SystemManager` class also now checks for screaming snake case in environment variables and lowercase dot separated in system properties

## Types of Changes

### DIFF
--- a/java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
@@ -1,48 +1,38 @@
 package com.saucelabs.saucebindings;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.saucelabs.saucebindings.citools.Bamboo;
+import com.saucelabs.saucebindings.citools.CITool;
+import com.saucelabs.saucebindings.citools.CircleCI;
+import com.saucelabs.saucebindings.citools.DefaultTool;
+import com.saucelabs.saucebindings.citools.GitHub;
+import com.saucelabs.saucebindings.citools.GitLab;
+import com.saucelabs.saucebindings.citools.Jenkins;
+import com.saucelabs.saucebindings.citools.TeamCity;
+import com.saucelabs.saucebindings.citools.TravisCI;
 import java.util.Map;
+import java.util.function.Supplier;
 
 /**
  * Determines the information a test needs based on what CI Tool is being used. It does not need to
  * be initialized.
  */
 public abstract class CITools {
-  private static String ciToolName;
+  private static CITool ciTool;
 
   /** Map of CI Tools and what Environment variable, if present, indicates its usage. */
-  public static final Map<String, String> KNOWN_TOOLS =
+  public static final Map<String, Supplier<CITool>> KNOWN_TOOLS =
       ImmutableMap.of(
-          "Bamboo", "bamboo_agentId",
-          "Circle", "CIRCLE_JOB",
-          "Github", "GITHUB_SHA",
-          "GitLab", "CI",
-          "Jenkins", "JENKINS_HOME",
-          "TeamCity", "TEAMCITY_PROJECT_NAME",
-          "Travis", "TRAVIS_JOB_ID");
-
-  /** The Build Name and Build Number values to properly differentiate test runs. */
-  public static final Map<String, ImmutableList<String>> BUILD_VALUES =
-      ImmutableMap.of(
-          "Bamboo", ImmutableList.of("bamboo_shortJobName", "bamboo_buildNumber"),
-          "Circle", ImmutableList.of("CIRCLE_JOB", "CIRCLE_BUILD_NUM"),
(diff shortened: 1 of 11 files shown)


java/main/src/main/java/com/saucelabs/saucebindings/citools/DefaultTool.java
```diff
@@ -0,0 +1,24 @@
+package com.saucelabs.saucebindings.citools;
+
+import com.saucelabs.saucebindings.SystemManager;
+
+public class DefaultTool implements CITool {
+
+  @Override
+  public String getToolName() {
+    String toolName = SystemManager.get("SAUCE_TOOL_NAME");
```
diemol: What is `SAUCE_TOOL_NAME`? Do you mean the CI tool used?

java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
```diff
@@ -64,10 +50,23 @@ public static String getBuildName() {
...
-    if (!getCiToolName().equalsIgnoreCase("unknown")) {
-      String buildNumberKey = BUILD_VALUES.get(getCiToolName()).get(1);
-      return SystemManager.get(buildNumberKey);
+    return getCiTool().getBuildNumber();
+  }
+
+  private static CITool getCiTool() {
+    if (ciTool != null) {
+      return ciTool;
+    } else if (SystemManager.get("SAUCE_BUILD_NAME") != null) {
```
diemol: Why do you only check for `SAUCE_BUILD_NAME` and not the other two?

diemol (changes requested):
Overall, this looks good. I am unsure if breaking the CI tools apart like this will be helpful in the future, but I am not against it.

My major concern is that getting the CI tool information is vital for Insights, therefore I would expect to see tests in this PR as well.

One thing to add is that this change relies on some environment variables, either present on a known CI, or some environment variables the user can set. However, I don't see any documentation related to that. I would love to see code changes going together with changes in the docs. I don't expect @nadvolod to read all PRs and document them.

java/main/src/main/java/com/saucelabs/saucebindings/citools/DefaultTool.java
titusfortner: It's a unique key for the user to set a custom CI tool name if they want. Just needs something to fit the pattern. There might be a better name to make it more clear?

java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
titusfortner: You can get away with the defaults for the other 2, but if Build Name isn't set, there's no reason to track tool name or build number

titusfortner:
>  I am unsure if breaking the CI tools apart like this will be helpful in the future,

The main reason I went through this additional trouble is that GitHub doesn't give me sufficiently useful values based on individual environment variables, so I wanted a way to combine several of them.

At the point we do that, we can (should?) allow a user to override the defaults and specify exactly what they want.

As a history note: we no longer want people to use Sauce Bindings without one of the Test Runner libraries. There just aren't good enough benefits to make it worth it. So things you can do (e.g., set custom test names or build names) with Bindings directly, can't be done when using it with a Test Runner library. This makes previous design choices more limiting than originally intended.

So for documenting this... it didn't used to matter, because it's a good default that is much better than nothing if you didn't pass something in. Now it matters a little because we effectively removed the user ability to customize these values directly.

java/main/src/main/java/com/saucelabs/saucebindings/citools/DefaultTool.java
titusfortner: I'm changing this to `SAUCE_CLIENT_PLATFORM`. alternatives I considered - "executing machine" "client machine" "executing platform" "user machine" "user platform"
Let me know if there is another one to prefer. 

This is really only to override the default detection of CI Tool, or "opt-out" of sending the information

java/main/src/main/java/com/saucelabs/saucebindings/CITools.java
titusfortner: No, this is a good point. I updated all the classes to check for defaults for everything.

titusfortner:
This has been updated based on provided feedback

diemol (approved):

PR #340, opened 2024-10-18
[java] implement session methods with SauceREST library

titusfortner:
# One-line summary

Use REST API where possible

## Description
REST API should be preferred over sending commands via execute script to the driver
It is especially useful when stopping the session if the driver is corrupted or inaccessible. This should reduce 90 second timeout errors
Mostly this is internal implementation details.

Obviously there is a lot more we can do here down the road, but here are the new user facing methods...
```
// Assets
session.assets.listAssets()
session.assets.downloadAll(Path path)
session.assets.download(Path path, TestAsset asset)

// Update after session start
session.changeBuildName(String name)
session.changeTestName(String name)
session.AddTag(String tag)
session.AddCustomData(String key, Object value)

// Getters
session.getTags()
session.getCustomData()

// Session Change
session.delete()
```

## Types of Changes

## Considerations
I'm mostly just looking for feedback on this implementation and the general idea. This wasn't as hard to implement as I was concerned with, and there's no reason we have to implement everything all at once.

### DIFF
--- a/java/main/src/main/java/com/saucelabs/saucebindings/SauceRest.java
+++ b/java/main/src/main/java/com/saucelabs/saucebindings/SauceRest.java
@@ -1,0 +1,98 @@
+package com.saucelabs.saucebindings;
+
+import com.saucelabs.saucerest.SauceREST;
+import com.saucelabs.saucerest.TestAsset;
+import com.saucelabs.saucerest.api.JobsEndpoint;
+import com.saucelabs.saucerest.model.jobs.Job;
+import com.saucelabs.saucerest.model.jobs.JobAssets;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import lombok.SneakyThrows;
+import org.openqa.selenium.remote.SessionId;
+
+public class SauceRest {
+
+  private final String id;
+  private final JobsEndpoint jobs;
+
+  public SauceRest(DataCenter dataCenter, SessionId id) {
+    this.id = id.toString();
+    SauceREST rest =
+        new SauceREST(com.saucelabs.saucerest.DataCenter.fromString(dataCenter.name()));
+    this.jobs = rest.getJobsEndpoint();
+  }
+
+  @SneakyThrows
+  public void setResult(Boolean result) {
+    jobs.changeResults(id, result);
+  }
+
+  public Map<String, String> getCustomData() {
+    return getDetails().customData;
+  }
+
+  @SneakyThrows
+  public void addCustomData(Map<String, String> data) {
+    jobs.addCustomData(id, data);
+  }
+
+  public List<String> getTags() {
+    return getDetails().tags;
+  }
+
+  @SneakyThrows
+  public void addTags(List<String> tags) {
+    jobs.addTags(id, tags);
+  }
+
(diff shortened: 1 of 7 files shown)


diemol (commented):
I think this makes sense, it helps the user. 

For us, this would potentially make us maintainers of SauceREST.

titusfortner:
> this would potentially make us maintainers of SauceREST

Yes; to me it makes sense for SauceREST, Sauce Jenkins Plugin and Sauce Bindings to all be actively maintained and kept in sync, so we just need to figure out the organization/prioritization/time requirements.

java/main/src/main/java/com/saucelabs/saucebindings/SauceRest.java
```diff
@@ -0,0 +1,88 @@
...
+import lombok.SneakyThrows;
+import org.openqa.selenium.remote.SessionId;
+
+public class SauceRest {
+
+  private final String id;
+  private final JobsEndpoint jobs;
+
+  public SauceRest(DataCenter dataCenter, SessionId id) {
+    this.id = id.toString();
```
titusfortner, line 20: Add null check here

java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
```diff
@@ -49,6 +56,8 @@ public RemoteWebDriver start() {
     }
 
     this.driver = createRemoteWebDriver(getSauceUrl(), sauceOptions.toCapabilities());
+    this.id = driver.getSessionId();
+    this.rest = new SauceRest(dataCenter, id);
```
titusfortner: use `this.id`

titusfortner:
* Use `update` instead of `change`
* Add getters for test name and build name

java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
```diff
@@ -334,8 +385,15 @@ private void validateSessionStarted(String method) {
   }
 
   private void quit() {
-    driver.quit();
-    driver = null;
+    try {
+      driver.quit();
+    } catch (WebDriverException e) {
```
titusfortner: This probably can be more than just WebDriverException. We really want the session to stop if we use this method, and this is a backup to sending the quit command. We do want this because if the driver is working, we want the quit command in the logs with timestamp, which you don't get if you just call stop via API.

titusfortner:
I had to rebase off main, so might be a few things to tidy up

java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
titusfortner: I changed this to Exception

java/main/src/main/java/com/saucelabs/saucebindings/SauceSession.java
titusfortner: this was not the problem. the problem was the mocks

titusfortner:
I'm keeping `change*()` instead of a more idiomatic `update*()` because it matches SauceREST. Let's all be wrong together.

PR #341, opened 2024-11-04
[plugin] initial working implementation of IntelliJ plugin

titusfortner:
This is updated to the new environment variable from #342 so that needs to be merged first. Putting this in draft until then.

Very rough and basic proof of concept implementation of an intellij plugin to run tests on Sauce Labs


diemol (commented):
Since we want to change the name of the env var in the other PR, that is the only change I would do to this one. Update the name of the env var.

PR #342, opened 2024-11-04
[java] set isDisabled to true by default (make Bindings opt-in)

titusfortner:
I think we want to make Sauce Bindings opt-in instead of opt-out.

This has the advantage of not changing default behavior when registering extension or adding rule.
This would require the user to update their CI tool to run with an environment variable or system property.
This would require the user to add `static { System.setProperty("sauce.disabled", "false") }` in order to run tests on Sauce Labs.
If we have this logic in CI plugins (e.g. Jenkins), it would just work on CI. If we create IDE plugins, we just add the toggle to the Run configs or context clicks.

Now that we are making this switch, do we want to make this `sauce.enabled` instead? Probably.

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -61,6 +61,9 @@ jobs:
         run: git --no-pager diff --exit-code
   main:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        profile: [enabled, disabled]
     steps:
       - uses: actions/checkout@v4
       - name: Set up JDK 11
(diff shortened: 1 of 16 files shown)


diemol (commented):
Yeah, I would change the name to enabled.

titusfortner:
Ok, so part of making things opt-in...
I switched from maven executions to profiles and updated the Github actions workflows accordingly.

I added the convenience method for enabling the extension rather than requiring a system property, but added the details in the code comments.

Really when it comes down to it, I want to make sure that these examples are what we want users to do:
https://github.com/saucelabs/sauce_bindings/tree/513cd55a16b33c73e104b9011bdef05075518371/java/junit5/src/test/java/com/saucelabs/saucebindings/junit5/examples

diemol (approved):
I'd say this is good enough.

